Issue #30, opened 2015-05-11
fix warnings


rgetz:
I added a branch "fix_warnings", which turns on a few more warnings in gcc.

When people want to take a break, and mindless fix some of the unsigned/signed comparisons issues -- that is the place to do it.

It's mostly in the plugsin now.

-Robin

rgetz:
This is been merged, and the branch removed.

Thanks

PR #34, opened 2015-05-27
Fix warnings

dNechita:
There are no more warnings to be shown when compiling the project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,10 @@ endif
 
 CFLAGS := $(shell $(PKG_CONFIG) --cflags $(DEPENDENCIES)) \
 	-I$(SYSROOT)/usr/include $(if $(WITH_MINGW),-mwindows,-fPIC) \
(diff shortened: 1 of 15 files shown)


dNechita:
These changes are for enhancement issue: #30.

rgetz:
Now that there are no warnings, did we want to add -Werror? To make sure we don't introduce any others?

dNechita:
Yes, that makes sense. I'll add the -Werror flag directly to the master but after the pull request gets merged.

rgetz:
So, we are just waiting on an ack from Paul before this gets merged into master?

-Robin

pcercuei:
Looks OK to me, but please don't add -Werror, otherwise the OSC will never build with MinGW.

dNechita:
Paul removed all warnings that come along with a MinGW compilation so I've added the -Werror flag.

-Dan

PR #41, opened 2015-06-24
Add support for tuning the AD9361 against an external frequency counter via SCPI

radhermit:
This work isn't fully completed/cleaned/squashed yet (so please don't merge it quite yet), but it seems to be most of the way there so I thought it would be good enough to post for an initial review.

Really the SCPI pieces need re-architecting especially if we want to support more functionality without having to hardcode in the proper sequences of commands for every type of wanted query, but I didn't get around to that with this work.


rgetz:
I'm not sure what you mean by:
>Really the SCPI pieces need re-
>architecting especially if we want to 
>support more functionality without 
>having to hardcode in the proper 
>sequences of commands for every type 
>of wanted query

? 

Since every query on every instrument is different, and needs the goofy scanf/printf string processing - I'm not sure how else to architect things?

Michael can comment on if he would prefer the button in the advanced tab. 

Can we add some more docs to:

http://wiki.analog.com/resources/eval/user-guides/ad-fmcomms2-ebz/hardware/tuning

On the process and results?



- Robin

radhermit:
On 2015-06-25 00:44, rgetz wrote:
> I'm not sure what you mean by:
> >Really the SCPI pieces need re- architecting especially if we want to
> >support more functionality without having to hardcode in the proper
> >sequences of commands for every type of wanted query

> ?

> Since every query on every instrument is different, and needs the
> goofy scanf/printf string processing - I'm not sure how else to
> architect things?

Oh I know, that was mostly just me reiterating my complaint that doing
glorified scripting and string processing would be cleaner/simpler in
some other language. :)

> Michael can comment on if he would prefer the button in the advanced
> tab.

I put it in the regular tab mostly to more easily show the tuning values
changing to the user instead of just showing a progress bar, but most people
probably don't care about seeing that.

Moving it to the advanced tab would just require accessing most of the
device values from sysfs using libiio.

> Can we add some more docs to:

> http://wiki.analog.com/resources/eval/user-guides/ad-fmcomms2-ebz/hardware/tuning

> On the process and results?

Yep, I can look into that.

radhermit:
I wrote some initial docs at http://wiki.analog.com/resources/eval/user-guides/ad-fmcomms2-ebz/hardware/tuning.

Also, this is probably decent enough to be reviewed and merged now.

rgetz:
I see a save to EEPROM. Is there a "load from EEPROM"?

radhermit:
Ah, I forgot it. I'll add it now.

rgetz:
OK, I have not looked at the code - I will leave that for Michael/Dan/Paul.

I would assume that the load/save buttons are always there (if you are root, and it can find the fru-dump utility). If SCPI is not working, then the cal method, and cal button are hidden? (Does that make sense to everyone?)

The only adds to the docs that I can think of is to make sure (a) you are running as root, (b) SCPI is loaded, and connected (you don't say how to do that). 

Not to be nit-picky or anything - but since the window manager you use is different than the ones running on the target - the screen shots look different.

Based on a discussion I just had with Michael - I also need to add an "impact" section in the top.

rgetz:
I updated the wiki page with the section I wanted to add, and move things around a bit (changed the titles).

radhermit:
On 2015-07-01 04:40, rgetz wrote:
> OK, I have not looked at the code - I will leave that for
> Michael/Dan/Paul.

Just to note, I have a couple more things to push today as well.

> I would assume that the load/save buttons are always there (if you are
> root, and it can find the fru-dump utility). If SCPI is not working,
> then the cal method, and cal button are hidden? (Does that make sense
> to everyone?)

I disable them instead of hiding them. If we want to hide them, it would
make more sense to restructure how they're added to the layout so it
wouldn't look as funny.

> The only adds to the docs that I can think of is to make sure (a) you
> are running as root, (b) SCPI is loaded, and connected (you don't say
> how to do that). 

I'll add that now.

> Not to be nit-picky or anything - but since the window manager you use
> is different than the ones running on the target - the screen shots
> look different.

Hah, I can regrab them from the zedboard environment later today, I've
been running stuff mostly remotely since it's easier.

rgetz:
>I disable them instead of hiding them. 

Totally up to Dan/Paul/Michael on how they want to manage things.

-Robin

radhermit:
On 2015-07-01 12:16, rgetz wrote:
> >I disable them instead of hiding them. 

> Totally up to Dan/Paul/Michael on how they want to manage things.

The main issue with hiding things is where to stick that seemingly one
off FMComms3 up/down converter button. It could be reflowed to the left
but that starts getting unnecessarily fiddly I think.

mhennerich:
It's ok to leave them on the main plugin - 
But please hide them - that's what we normally do.

-Michael

radhermit:
On 2015-07-01 15:31, Michael Hennerich wrote:
> It's ok to leave them on the main plugin - 
> But please hide them - that's what we normally do.

Already did that this morning, just need to push it with other things.

pcercuei:
Let me check if it still builds with MinGW / works in Windows.

pcercuei:
Build fails: https://gist.github.com/pcercuei/ba24928665588c9ad725

mhennerich:
Added some comments in the patch series.

rgetz:
With respect to: Build fails: https://gist.github.com/pcercuei/ba24928665588c9ad725

Did we just want to make eeprom.c not compile on ming? (EEPROM is only going to work on local targets for now).

mhennerich:
Makes sense. 
I would suspect if an instrument follows the SCPI spec. 
The generic strings should just work.

-Michael

mhennerich:
send '*IDN?',
received 'PHILIPS, PM6681, 0, MAIN V1.09  26 JAN 2001 / GPIB V1.13  26 JAN 2001

send ':MEAS:FREQ?',
received '+1.018836782E+07
'
send ':MEAS:FREQ?',
received '+1.018843874E+07
'
send ':READ?',
received '+1.018835849E+07
'
send ':READ?',
received '+1.018834220E+07
'

radhermit:
On 2015-07-02 10:47, Michael Hennerich wrote:
> send '*IDN?',
> received 'PHILIPS, PM6681, 0, MAIN V1.09  26 JAN 2001 / GPIB V1.13  26 JAN 2001

> send ':MEAS:FREQ?',
> received '+1.018836782E+07
> '
> send ':MEAS:FREQ?',
> received '+1.018843874E+07
> '
> send ':READ?',
> received '+1.018835849E+07
> '
> send ':READ?',
> received '+1.018834220E+07
> '

Yep, ':MEAS:FREQ?', was the first thing I tried to query the HP model
but it never returns anything so I went with READ instead. Perhaps I
don't set some other initialization fields that measure expects.

rgetz:
Yeah, what I found is "SCPI" isn't a spec, as much of an intent.

Everything is close, but not quite. That's why I started to make the instrument specific stuff...

Rather than string comparisons in the parts that do the work -- we could implement our own fops to make it easier to do various things on different pieces of equipment. (not sure if that would make it easier, or just add extra complexity).

Also a few enums to help define the different equipment might be good as well...

-Robin

PR #42, opened 2015-06-29
plugins/ad9361_multichip_sync: Factor out MCS so it can be used outside

mhennerich:
Factor out MCS so it can be used outside OSC as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ endif
 
 OSC_OBJS := osc.o oscplot.o datatypes.o int_fft.o iio_widget.o fru.o dialogs.o \
 	trigger_dialog.o xml_utils.o libini/libini.o libini2.o phone_home.o \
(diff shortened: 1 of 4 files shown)


rgetz:
I think you removed this line:
plugins/fmcomms2_adv.c: 740 `-	for (step = 0; step <= 5; step++) {`

and replaced it with this one:
plugins/ad9361_multichip_sync.c:53 `+	for (step = 1; step <= 5; step++) {`

I think that adds back a bug you just fixed. (should start from zero, shouldn't it?)
https://ez.analog.com/message/201231

-Robin

mhennerich:
https://github.com/analogdevicesinc/linux/commit/69db93dfd207f38c58b47b0bde46624bac664599

With this commit I fixed it on the kernel side.
But you're right we should keep the 0..5 for a while.

-Michael

PR #47, opened 2015-07-10
Fix filters path for non-standard prefixes

pcercuei:
This pull request does a few things:
* move the load_fir_filter( ) function to osc.c, as it was duplicated for the fmcomms2 and fmcomms5 plugins,
* Handle the @FILTERS@ token when loading a profile, which will be translated to the system installation's filter path,
* Use the @FILTERS@ in all the profiles that load a FIR filter, so that those profiles will correctly load independently of the prefix used during the compilation.
* As a result, this fixes the profiles not loading the FIR filter file under Windows.


rgetz:
> move the load_fir_filter( ) function to osc.c, as it was duplicated for the
> fmcomms2 and fmcomms5 plugins,

I get we should use a common thing - but in the osc.c, not something new in ./plugins/ ?

pcercuei:
Should I move it to a separate file in plugins/ ?
I don't know what's Dan's policy on this.

dNechita:
You can look at the dac_data_manager.h. It's in the plugins/ and it's common code for all plugins that contain a DDS/DAC Output functionality.

PR #48, opened 2015-07-24
Add test support for DCXO calibration and saving to eeprom for fmcomms2 and fmcomms5

radhermit:
A couple things to note. First off, I think there should be a better method than how I'm destroying the plot in 1d87e15 but that forces the markers to work correctly when doing multiple calibration runs for fmcomms5 boards.

Secondly, adding variable amounts of wait after toggling channels on/off and then measuring the rssi/gain feels somewhat wrong. Especially since even with a few seconds pause the tests still fail, albeit much more rarely now. Any suggestions to improve that would be appreciated.

Anyway, with this fmcomms2 tests pass and push calibrated coarse/fine tuning values into the eeprom while fmcomms5 boards with adf5355 also push their clock frequency into the eeprom while old fmcomms5 boards without the adf5355 should timeout during the initial calibration and thus do not pass the tests.

Also, please try the windows build against this. I think I handled everything that won't work, but I could have missed something, thanks in advance.


rgetz:
>Secondly, adding variable amounts of 
>wait after toggling channels on/off and 
>then measuring the rssi/gain feels 
>somewhat wrong

I think this is because the act of writing things to sysfs just queues up SPI transactions. I don't think we wait for the SPI to be drained (ie actually hit the device) before we do the read. If use space is fast enough, I don't know if the reads always use a fresh read, or if if stashes things. 

Michael would know the details n

mhennerich:
All API transactions via the regular sysfs API are mutual exclusive and blocking. They won't return unless the command sequence hit successfully the underlying device. Ini processing was different, but I thought be added the option to revert to the original sequential processing? In addition interactions via debugfs are also quite different.

Can you point me to the sequence of things that are not working ?

-Michael

rgetz:
I think the system is something like:
(With loop back cables):
Turn on DDS
Set Rx and Tx LO to be same
Set gain mode
Wait
Measure RSSI

Why does the wait need to be there?

mhennerich:
What gain mode?
There are various events to trigger the RSSI algorithm. IIRC our default for slow attack is "gain change occurs". For a gain change the gain update counter needs to expire. Our default gain update interval is 1ms and RSSI duration is also 1ms.

These should be the delays.

-Michael

pcercuei:
It needs a #ifndef _WIN32 to protect the inclusion of scpi.h in plugins/fmcomms5.c, otherwise the MinGW build will fail because of -Werror. With that small fix it builds fine.

dNechita:
>First off, I think there should be a better method than how I'm destroying the plot in  1d87e15  but >that forces the markers to work correctly when doing multiple calibration runs for fmcomms5 boards.

It is the right way to destroy the plot. I'm using the same method:
https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/oscplot.c#L6059

radhermit:
On 2015-07-27 08:49, Dan Nechita wrote:
>  >First off, I think there should be a better method than how I'm
>  >destroying the plot in  1d87e15  but >that forces the markers to
>  >work correctly when doing multiple calibration runs for fmcomms5
>  >boards.

> It is the right way to destroy the plot. I'm using the same method:
> https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/oscplot.c#L6059

I was more alluding to the idea that perhaps plots should be destroyed
automatically on leaving a section during sequential ini loading for
tests, but I guess that might be hard to get right without confusing
people that just want to stop/start the same plot in different sections
if that's necessary.

radhermit:
On 2015-07-27 07:35, Paul Cercueil wrote:
> It needs a #ifndef _WIN32 to protect the inclusion of scpi.h in
> plugins/fmcomms5.c, otherwise the MinGW build will fail because of
> -Werror. With that small fix it builds fine.

Thanks, the windows build should now work.

PR #50, opened 2015-09-04
Create adi_update_tools.sh

tinglic:


pcercuei:
Hi,

It makes little sense to have the Debian package of the iio-oscilloscope provide this file, for several reasons:
* it is not specific to the iio-oscilloscope,
* it will compile and install libraries and programs that will overwrite the currently installed ones. This include overwriting the files that would be installed by the iio-oscilloscope Debian package.

Besides, this commit just adds the file under the debian/ directory, it won't be installed automatically by the generated .deb unless you modify the config files as well.

-Paul

radhermit:
Closing since adding this to the repo doesn't make much sense due to the reasons mentioned previously.

PR #56, opened 2017-01-04
Changed in_voltage_quadrature_tracking_en

ttyler777:
in_voltage_quadrature_tracking_en does not exist as an attribute.  Changed to quadrature_tracking_en.


mhennerich:
Looks good - merged.
Thanks for pointing out.

-Michael

PR #63, opened 2018-01-02
increase buffer size to prevent "plugins/debug.c:1175:30: error: ‘__b…

njpillitteri:
…uiltin___snprintf_chk’ output may be truncated before the last format character [-Werror=format-truncation=]"


mhennerich:
Applied - Thanks for the contribution.

PR #66, opened 2018-02-09
build: fix & update dependency checks

commodo:
* iterate over all dependencies to find out which one is missing ; pkg-config is a all-or-nothing method ; so if any dependency is missing, it will always point to glib as not being found
* dependency check is the first build step ; this has been made sure to run as first even for multi-job builds
* Added checks for `matio` and `libad9361-iio` to dep check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ PKG_CONFIG_PATH := $(SYSROOT)/usr/share/pkgconfig:$(SYSROOT)/usr/lib/pkgconfig:$
 PKG_CONFIG := env PKG_CONFIG_SYSROOT_DIR="$(SYSROOT)" \
 	PKG_CONFIG_PATH="$(PKG_CONFIG_PATH)" pkg-config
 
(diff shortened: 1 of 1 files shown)


lclausen-adi:
Should we just abort the build if any of the dependencies are missing?

lclausen-adi:
That's what I meant. Make it explicit that the build will fail because dependency X was not found. Even before starting the build.

commodo:
Changelog v1 -> v2:
* iterate over all dependencies to find out which one is missing ; pkg-config is a all-or-nothing method ; so if any dependency is missing, it will always point to glib as not being found
* dependency check is the first build step ; this has been made sure to run as first even for multi-job builds
* Added checks for `matio` and `libad9361-iio` to dep check
* update PR comment

commodo:
> I'm OK with this, but I feel like there must be a better way than having the check as a dependency for all rules. But for now its OK, just keep your eyes open in case you find a better way.

agreed ;
using CMake or Autotools would be the general good approach ;
we later could do an overhaul to CMake ;
this is more of a short-fix until we get to that point

Issue #72, opened 2018-03-17
LDFLAGS:  -L$(SYSROOT)/usr/lib vs  -L$(SYSROOT)/usr/lib64


UweBonnes:
Hello,

compiling on Opensuse Leap42.3 gives a lot of errors like
  LD      libosc.so
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: inkompatibles /usr/lib/libfontconfig.so wird bei der Suche nach -lfontconfig übersprungen
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: inkompatibles /usr/lib/libfreetype.so wird bei der Suche nach -lfreetype übersprungen

Changing
```
diff --git a/Makefile b/Makefile
index 418af2e..e08900f 100644
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ $(foreach dep,$(DEPENDENCIES),$(eval $(call dep_flags,$(dep))))
 
 LDFLAGS := $(DEP_LDFLAGS) \
        $(if $(WITH_MINGW),-lwinpthread) \
-       -L$(SYSROOT)/usr/lib -lmatio -lz -lm -lad9361
+       -L$(SYSROOT)/usr/lib64 -lmatio -lz -lm -lad9361
 
 ifeq ($(WITH_MINGW),y)
        LDFLAGS += -Wl,--subsystem,windows
```
makes the error go away for me, but a more general solution is needed.

commodo:
One generic way to fix this, is to add `-L$(SYSROOT)/usr/lib64` before `-L$(SYSROOT)/usr/lib`.
It's not perfect, but it should handle most cases.
If the lib is found in `-L$(SYSROOT)/usr/lib64`, it will be used.
If not, and is found in ``-L$(SYSROOT)/usr/lib`, it will be used.

It's also a simple solution to the problem.
I'm not sure about any other solution here.

I'm open to suggestions.
Otherwise, I propose we solve it like this.

I will spin up a PR with this change.

commodo:
@UweBonnes
Could you test the change in PR #77 ?
If you give the ok, we could then merge it.

Thanks :)
Alex

UweBonnes:
PR#77 fixes the problem on OpenSuse 42,3, X64

PR #77, opened 2018-03-28
 build: organize linker search paths in LDFLAGS

commodo:
Fixes: https://github.com/analogdevicesinc/iio-oscilloscope/issues/72
Fixes: https://github.com/analogdevicesinc/iio-oscilloscope/issues/79
    
Order/preference is: /usr/lib64, /usr/lib, /usr/lib32.
Suggestion came from Edward Kigwana on Github issue #79.
This should handle most standard cases/setups.
For the exotic cases, there's usually not much we can do anyway.
    
Handling multi-arch setups is not always straightforward on all distros and
the simplest way is to organize linker search paths, so that one has a
preference over the other.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,10 @@ $(foreach dep,$(DEPENDENCIES),$(eval $(call dep_flags,$(dep))))
 
 LDFLAGS := $(DEP_LDFLAGS) \
 	$(if $(WITH_MINGW),-lwinpthread) \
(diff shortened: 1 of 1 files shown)


commodo:
Waiting for confirmation from @UweBonnes on this one.

UweBonnes:
This solves the problem on both OpenSuse Leap 43 X64.

commodo:
if no objections, will merge this tomorrow

PR #80, opened 2018-04-09
oscplot.c: fix/add omitted break statements

commodo:
These were found by enabling some extra checks for the compiler.
They are small bugs that should be fixed.


commodo:
if no other objections, i'll merge this later today

Issue #71, opened 2018-03-17
Incomplete function declaration scpi_rx_setup()


UweBonnes:
Hello,

compiling master gives:
  LD      plugins/scpi.so
plugins/scpi.c: In function ‘scpi_handle’:
plugins/scpi.c:984:4: error: call to function ‘scpi_rx_setup’ without a real prototype [-Werror=unprototyped-calls]
    scpi_rx_setup(current_instrument);
    ^
plugins/scpi.c:601:6: note: ‘scpi_rx_setup’ was declared here
 void scpi_rx_setup()
      ^
cc1: all warnings being treated as errors
Makefile:140: recipe for target 'plugins/scpi.so' failed

I can fix e.g. by
diff --git a/plugins/scpi.c b/plugins/scpi.c
index f7ea719..bb1a547 100644
--- a/plugins/scpi.c
+++ b/plugins/scpi.c
@@ -598,7 +598,7 @@ void scpi_rx_set_bandwith_auto(double ratio)
 
 }
 
-void scpi_rx_setup()
+void scpi_rx_setup(void *bla)
 {
        static time_t rx_cal_time = 0;

Bye

rgetz:
Uwe:

What toolchain are you using? 

gcc will warn on your patch due to -Wunused-parameter , which is set in -Wextra. 

?

-Robin

UweBonnes:
I use OpenSuse Leap 42.3. The default GCC is gcc version is 4.8.5 (SUSE Linux). However "void scpi_rx_setup();" is no valid C code as far as I know.

UweBonnes:
The diff I gave above was only to menat as a hint. A more complete solution would be:
diff --git a/plugins/scpi.c b/plugins/scpi.c
index f7ea719..6455611 100644
--- a/plugins/scpi.c
+++ b/plugins/scpi.c
@@ -598,10 +598,10 @@ void scpi_rx_set_bandwith_auto(double ratio)
 
 }
 
-void scpi_rx_setup()
+void scpi_rx_setup(struct scpi_instrument *instrument)
 {
        static time_t rx_cal_time = 0;
-
+        (void)instrument;
        scpi_fprintf(&spectrum_analyzer, ":DISP:TRACE:Y:RLEVEL %d DBM\n", 10);
        /* Turn averaging off */
        scpi_fprintf(&spectrum_analyzer, ":AVER OFF\n");

rgetz:
[...]

PR #78, opened 2018-03-29
build: fix some compiler reported warnings/errors

commodo:
This PR enables some extra compiler checks.
They build fine on Ubuntu 17.10 with GCC 7.2.0.

Also, fixes https://github.com/analogdevicesinc/iio-oscilloscope/issues/71 .

Signed-off-by: Alexandru Ardelean <alexandru.ardelean@analog.com>

PR #81, opened 2018-04-09
plugins/scpi: remove argument in scpi_rx_setup() call

commodo:
Fixes https://github.com/analogdevicesinc/iio-oscilloscope/issues/71

The `scpi_rx_setup()` function does not take any args, but it's called
with one.
The compiler does not seem to complain, especially when checks are a bit lax.
And even though this is not a problem (functionally), it is incorrect (a
bug). This fixes it.

### DIFF
--- a/plugins/scpi.c
+++ b/plugins/scpi.c
@@ -981,7 +981,7 @@ static int scpi_handle(int line, const char *attrib, const char *value)
 		}
 	} else { /* current_instrument == &spectrum_analyzer */
 		if (MATCH_ATTRIB("setup")) {
-			scpi_rx_setup(current_instrument);
+			scpi_rx_setup();
 		} else if (MATCH_ATTRIB("center")) {
 			scpi_rx_set_center_frequency(atoll(value));
 		} else if (MATCH_ATTRIB("span")) {


lclausen-adi:
Please also add (void) to scpi_rx_setup().

commodo:
curios: just for `scpi_rx_setup()` ?
I have this change, which updates the complete `scpi.c` file for `(void)`

commodo:
sorry;
i meant, this change https://github.com/analogdevicesinc/iio-oscilloscope/pull/78/commits/d3171779d789595c3f9eaad0803ebbfe0e3c35c1

lclausen-adi:
ok, in that case all is good.

commodo:
@rgetz 
I took a closer look at the code.

The problem is that some `scpi` functions are used in the `fmcomms1` plugin.
`scpi_rx_setup()` is called without an instrument parameter in that plugin.

AFAICT, the `scpi_rx_setup()` function assumes the use of the `&spectrum_analyzer` instrument (for the RX path) which is a private attribute of the `scpi` plugin.
The `fmcomms1` plugin relies on using this (spectrum_analyzer) instrument via accessor functions, and not knowing about it.

In order to add the instrument as a param to `scpi_rx_setup()` would mean to export the type definition of the `scpi_instrument` outside of this plugin, which is a bit invasive.

There is a slight software design hiccup in the `scpi` plugin:
```
static int scpi_handle(int line, const char *attrib, const char *value)
{
        if (!strncmp(attrib, "rx.", sizeof("rx.") - 1))
                current_instrument = &spectrum_analyzer;
        else if (!strncmp(attrib, "tx.", sizeof("tx.") - 1))
                current_instrument = &signal_generator;
        else
                return -EINVAL;

        attrib += sizeof("rx.") - 1;

        if (MATCH_ATTRIB("serial")) {
             ......
        } else if (current_instrument == &signal_generator) {
             .....
        } else { /* current_instrument == &spectrum_analyzer */
                if (MATCH_ATTRIB("setup")) {
                        scpi_rx_setup(current_instrument);
```
Where, `scpi_rx_setup()` is given a reference to the `&spectrum_analizer` object via param and not using that param, but still initializing the same `&spectrum_analizer` object.

I'd still say, removing the `current_instrument` is the best fix (in the current scheme of things).
Something else would probably require a bigger rework than is currently worth having.

commodo:
re-opened PR;
@rgetz : any strong opinions against this approach ?  if not, i'd say we merge this as is, and see later ; though, my feeling si that scopy will replace this project eventually ;

commodo:
will merge this tomorrow if no objections

Issue #93, opened 2018-08-06
Build error


Skizzo69:
[skizzo@localhost iio-oscilloscope]$ make -j 4
  CC      osc.o
  CC      oscplot.o
  CC      int_fft.o
  CC      datatypes.o
  CC      iio_widget.o
  CC      fru.o
  CC      dialogs.o
  CC      trigger_dialog.o
osc.c: In function ‘capture_start’:
osc.c:1325:70: error: cast between incompatible function types from ‘gboolean (*)(void)’ {aka ‘int (*)(void)’} to ‘gboolean (*)(void *)’ {aka ‘int (*)(void *)’} [-Werror=cast-function-type]
   capture_function = g_timeout_add_full(G_PRIORITY_DEFAULT_IDLE, 50, (GSourceFunc) capture_process, NULL, NULL);
                                                                      ^
cc1: all warnings being treated as errors
make: *** [Makefile:145: osc.o] Error 1
make: *** Waiting for unfinished jobs....
In function ‘marker_set’,
    inlined from ‘set_marker_labels’ at oscplot.c:5469:4:
oscplot.c:5433:3: error: ‘strncpy’ output may be truncated copying 6 bytes from a string of length 127 [-Werror=stringop-truncation]
   strncpy(priv->markers[i].label, buf, 6);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘marker_set’,
    inlined from ‘set_marker_labels’ at oscplot.c:5460:4:
oscplot.c:5433:3: error: ‘strncpy’ output may be truncated copying 6 bytes from a string of length 127 [-Werror=stringop-truncation]
   strncpy(priv->markers[i].label, buf, 6);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘marker_set’,
    inlined from ‘set_marker_labels’ at oscplot.c:5453:4:
oscplot.c:5433:3: error: ‘strncpy’ output may be truncated copying 6 bytes from a string of length 127 [-Werror=stringop-truncation]
   strncpy(priv->markers[i].label, buf, 6);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make: *** [Makefile:143: oscplot.o] Error 1
[skizzo@localhost iio-oscilloscope]$ 

[...]

PR #94, opened 2018-08-07
Fix gcc8 warnings

lclausen-adi:
This should fix the warnings reported in #93 

Unfortunately I don't have gcc-8 so can't verify that this is enough to fix all warnings that are generated by gcc-8.

### DIFF
--- a/oscplot.c
+++ b/oscplot.c
@@ -5430,9 +5430,9 @@ static inline void marker_set(OscPlot *plot, int i, char *buf, bool force)
 		priv->markers[i].active = TRUE;
 
 	if (priv->markers[i].graph) {
-		strncpy(priv->markers[i].label, buf, 6);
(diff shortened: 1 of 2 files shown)


oscplot.c
```diff
@@ -5430,44 +5430,37 @@ static inline void marker_set(OscPlot *plot, int i, char *buf, bool force)
 		priv->markers[i].active = TRUE;
 
 	if (priv->markers[i].graph) {
-		strncpy(priv->markers[i].label, buf, 6);
+		snprintf(priv->markers[i].label, sizeof(priv->markers[i].label), buf, i);
```
commodo: would it be an idea to also const-ify the `char *buf` param ?

commodo:
rest looks good

oscplot.c
lclausen-adi: Yes, but that is a separate change.

PR #95, opened 2018-08-10
build: add basic Travis CI integration to the project

commodo:
Mostly targeting Linux & OS X.
On OS X version 10.11 there doesn't seem to be support for `clock_gettime()` [or I could not find it] so that version was skipped and only 10.12 was added.

Tested on Ubuntu 18.04 as well ; mostly that install paths work fine.
On OS X will need to do testing and iron out issues.

Most stuff in this Travis-CI integration has been loosely adapted from Scopy's Travis-CI integration.


commodo:
Forgot to add: build reference is here https://travis-ci.org/commodo/iio-oscilloscope/builds
Build 29 is the current one that works.

PR #97, opened 2018-08-23
serial_context: Add serial context option

mirceacaprioru:
This patch will add the possibility to connect to a serial device using
libiio serial context option. It will enable the use of UART IIO devices.

Example:
ad9361 no-Os driver connected to fpga with tinyiiod integrated
This will enable to test a no-Os driver similar to a linux iio driver and we can use iio-osciloscope to view data.


dialogs.c
```diff
@@ -338,6 +343,28 @@ static struct iio_context * get_context(Dialogs *data)
...
+		gchar *id = "serial:";
+		gchar *port = gtk_combo_box_get_active_text(
+				GTK_COMBO_BOX(dialogs.connect_seriald));
+		gchar *baud_rate = gtk_combo_box_get_active_text(
+				GTK_COMBO_BOX(dialogs.connect_serialbr));
+
+		port[strlen(port) - 1] = '\0';
+
+		gchar *result = malloc(strlen(id) + strlen(port) + strlen(baud_rate) + 2);
+		strcpy(result, id);
```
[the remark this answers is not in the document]
mhennerich: Or even asprintf()

mhennerich:
Doesn't compile on Linux, due to missing dependencies ...

/home/michael/devel/iio-oscilloscope/dialogs.c:27:61: fatal error: libserialport.h: No such file or directory
compilation terminated.

mirceacaprioru:
Changelog V1 -> V2
- replaced strcpy and strcat with snprintf
- freed allocated memory
- added libserialport to travis build

dialogs.c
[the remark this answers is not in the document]
lclausen-adi: asprintf() is not portable. But g_strdup_printf() would work.

CMakeLists.txt
```diff
@@ -92,6 +92,7 @@ pkg_check_modules(GTKDATABOX REQUIRED gtkdatabox)
 pkg_check_modules(FFTW3 REQUIRED fftw3)
 pkg_check_modules(LIBXML2 REQUIRED libxml-2.0)
 pkg_check_modules(LIBCURL REQUIRED libcurl)
+pkg_check_modules(LIBCURL REQUIRED libserialport)
```
mhennerich: typo ...LIBCURL REQUIRED libserialport

mirceacaprioru:
I like the idea of making it optional since there is also the issue that libserialport is not available for Ubuntu 14.04.

dialogs.c
```diff
@@ -436,6 +463,36 @@ static void refresh_usb(void)
...
+	/* clear everything, and scan again */
+	liststore = GTK_LIST_STORE(gtk_combo_box_get_model(GTK_COMBO_BOX(dialogs.connect_seriald)));
+	gtk_list_store_clear(liststore);
+
+	/*get serial ports*/
+	enum sp_return error = sp_list_ports(&ports);
+	if (error == SP_OK) {
+	for (i = 0; ports[i]; i++) {
+		buf = malloc(strlen(sp_get_port_name(ports[i])) + 5);
+		sprintf(buf, "%s ",  sp_get_port_name(ports[i]));
```
lclausen-adi: Just for my understanding, why the extra space at the end?

mirceacaprioru: This is a mistake from my part. Sorry for that,  it also eliminates the removal of that space when creating the URI. 
port[strlen(port) - 1] = '\0';

dialogs.c
```diff
@@ -1089,7 +1166,19 @@ void dialogs_init(GtkBuilder *builder)
 		}
 	}
 
+	/* Serial Backend - hide if not supported */
+#ifndef SERIAL_BACKEND
+	gtk_widget_hide(dialogs.connect_seriald);
+	gtk_widget_hide(dialogs.connect_serial);
+	gtk_widget_hide(dialogs.connect_serialbr);
+	gtk_widget_hide(GTK_WIDGET(gtk_builder_get_object(builder, "connect_serial_label")));
+#endif
```
commodo: you could move these under `refresh_serial()` when `ifndef SERIAL_BACKEND`

dialogs.c
```diff
@@ -338,6 +345,27 @@ static struct iio_context * get_context(Dialogs *data)
...
+		struct iio_context *ctx;
+		gchar *port = gtk_combo_box_get_active_text(
+				GTK_COMBO_BOX(dialogs.connect_seriald));
+		gchar *baud_rate = gtk_combo_box_get_active_text(
+				GTK_COMBO_BOX(dialogs.connect_serialbr));
+
+		/* Size is +3: for ':', ',' and '\0' */
+		int buf_size = sizeof("serial:") + strlen(port) + strlen(baud_rate) + 3;
+		gchar *result = malloc(buf_size);
+		snprintf(result, buf_size, "serial:%s,%s", port, baud_rate);
```
lclausen-adi: Please change this to [g_strdup_printf()](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strdup-printf) in a follow up patch.

PR #100, opened 2018-08-29
Code cleanups: part 3: implement cleanups Codacy & add cppcheck to Travis CI build

commodo:
This set of patches is the 3rd set in a series of cleanups.
It's all stuff that's been reported mostly by Codacy.

In this changeset, cppcheck is also being added to the Travis CI run.
Even though Codacy also uses cppcheck for some checks, it seems that output across cppcheck versions is not always consistent, so might as well run an instance in TravisCI which may catch stuff that are missed by the one from Codacy.
And if Codacy becomes unavailable, we still get the static analysis from this cppcheck.


plugins/scpi.c
```diff
@@ -309,7 +309,6 @@ static int tty_read(struct scpi_instrument *scpi)
 	clock_gettime(CLOCK_MONOTONIC, &ts_current);
 	nsecs = ts_current.tv_nsec + (SOCKETS_TIMEOUT * pow(10.0, 9));
 	ts_end.tv_sec = ts_current.tv_sec + (nsecs / pow(10.0, 9));
-	ts_end.tv_nsec = nsecs % (unsigned long long) pow(10.0, 9);
 	ts_end.tv_nsec = ts_current.tv_nsec;
```
lclausen-adi: But does that make sense? Shouldn't the second line not rather be `+=` instead of `=`?

commodo: I'm not sure to be honest.
I'll have to check

lclausen-adi:
Typo in https://github.com/analogdevicesinc/iio-oscilloscope/pull/100/commits/2e569955160038d150828e8e4c2e67a0a341c103 
s/their/they are/

commodo:
Changelog v1 -> v2:
* fixed commit description typo

plugins/scpi.c
commodo: yep, good point ; it makes sense to switch to `+=`
thanks for catching this

commodo:
Will merge next week if no objections

PR #101, opened 2018-09-03
oscplot.c: remove deprecated code

commodo:
Build fails with g_type_class_add_private() being marked as deprecated.
  https://travis-ci.org/analogdevicesinc/iio-oscilloscope/jobs/423795815

It fails on OS X, because libraries tend to be more up-to-date via
homebrew.

GLib provides a convenience macro to replace the g_type_class_add_private()
call. Then the G_TYPE_INSTANCE_GET_PRIVATE() needs to be replaced with
<type>_get_instance_private().

### DIFF
--- a/oscplot.c
+++ b/oscplot.c
@@ -400,7 +400,7 @@ struct _OscPlotPrivate
 	void *qcb_user_data;
 };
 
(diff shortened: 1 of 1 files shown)


commodo:
If no objections will merge tomorrow

PR #108, opened 2018-09-27
Replace snprintf() + fwrite() with fprintf()

lclausen-adi:
Replace the sequence of

	snprintf(buf, ...);
	fwrite(buf, ..., f);
with
	fprintf(f, ...);

This is semantically equivalent for the most part but removes the need for
a temporary buffer. This new solution differs in that the data written is
not truncated if it exceeds the size of the original temporary buffer.

Bulk of the transformation was done using the following coccinelle semantic
patch:

// <smpl>
@@
expression buf;
expression len;
expression f;
@@
+fprinf(f,
-snprintf(buf, len,
 ...);
-fwrite(buf, ..., f);
// </smpl>

Signed-off-by: Lars-Peter Clausen <lars@metafoo.de>

### DIFF
--- a/plugins/ad9371.c
+++ b/plugins/ad9371.c
@@ -1775,9 +1775,7 @@ static void ad9371_get_preferred_size(int *width, int *height)
 
 static void save_widgets_to_ini(FILE *f)
 {
-	char buf[0x1000];
-
-	snprintf(buf, sizeof(buf), "load_myk_profile_file = %s\n"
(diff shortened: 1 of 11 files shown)


commodo:
```
$ apt-get moo
                 (__) 
                 (oo) 
           /------\/ 
          / |    ||   
         *  /\---/\ 
            ~~   ~~   
..."Have you mooed today?"...
```

PR #114, opened 2018-10-19
dac_data_manager: Consider device buffer length alignment requirements

lclausen-adi:
DMA buffers typically have an alignment requirement for the buffer length.
Currently this alignment requirement is hardcoded for each platform in the
corresponding plugin.

Newer versions of the kernel will report the alignment requirements
associated with a device through the `length_align_bytes` property. If this
property is present read it and use its value instead of the hardcoded
value from the plugin.

This will allow the DAC data manager to adopt to changes in the alignment
requirements without having to update the plugin.

Signed-off-by: Lars-Peter Clausen <lars@metafoo.de>

### DIFF
--- a/plugins/dac_data_manager.c
+++ b/plugins/dac_data_manager.c
@@ -124,6 +124,7 @@ struct dac_data_manager {
 	unsigned dacs_count;
 	unsigned tones_count;
 	unsigned alignment;
+	bool hw_reported_alignment;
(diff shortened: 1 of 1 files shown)


commodo:
[LGTM](https://i.pinimg.com/564x/a1/4c/44/a14c441f4c65db137a3acedacd29d893.jpg)

PR #117, opened 2018-11-19
GCC8 Support

thekix:


plugins/ad9371.c
```diff
@@ -707,7 +707,7 @@ static void rssi_update_labels(void)
 	rssi_update_label(obs_rssi, "voltage2", false);
 }
 
-static gboolean update_display(void)
+static gboolean update_display(gpointer foo)
```
dNechita: Not a major issue but could 'foo' be replaced with 'unused' or 'unusedParam'? (or other naming similar to this)

thekix:
This patch series is not included in the master branch. Probably there is a problem with Travis. Should I do something?

commodo:
Well, I wanted to first make the Travis CI builds work.
There were some changes to libiio artifact generation that broke this build.

Hopefully, by next week, we should have the Travis CI builds fixed ; we're also building a Travis-CI pipeline, where libiio triggers libad9361-iio which then triggers iio-oscilloscope.
But, we can have this merged in master sooner than that.

PR #118, opened 2018-11-20
plugins: adrv9009: Add support for ADRV9008-1 and ADRV9008-2

mhennerich:


mhennerich:
This need to go into 2018_R2 and must be build into the Release image.

PR #121, opened 2019-01-30
build,travis: fix build & extend the build to cover a larger build matrix

commodo:
This changeset cover all builds that libiio supports, except for CentOS 6, which requires a newer glib2 library.
See commit `.travis.yml: drop support for CentOS 6`  for details.


commodo:
note: also dropping support for Xcode 8.0
see comment in commit `.travis.yml: drop build for Xcode 8 `

PR #123, opened 2019-02-06
Helpers and fixes

dNechita:
Adding a couple of helpers that will be required for the implementation of the new plugin which will support multiple adrv9009 devices.
I've also squeezed a couple of small fixes.

### DIFF
--- a/iio_widget.c
+++ b/iio_widget.c
@@ -30,6 +30,18 @@ void g_builder_connect_signal(GtkBuilder *builder, const gchar *name,
 		g_signal_connect(tmp, signal, callback, data);
 }
 
+void g_builder_connect_signal_data(GtkBuilder *builder, const gchar *name,
+	const gchar *signal, GCallback callback, gpointer data,
+	GClosureNotify destroy_data, GConnectFlags connect_flags)
+{
+	GObject *tmp;
+	tmp = gtk_builder_get_object(builder, name);
+	if (tmp == NULL)
+		fprintf(stderr, "Couldn't find object \"%s\".\n", name);
+	else
+		g_signal_connect_data(tmp, signal, callback, data, destroy_data, connect_flags);
+}
+
 void g_builder_bind_property(GtkBuilder *builder,
 	const gchar *source_name, const gchar *source_property,
 	const gchar *target_name, const gchar *target_property,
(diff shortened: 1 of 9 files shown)


iio_widget.c
```diff
@@ -77,6 +77,8 @@ static void iio_widget_init(struct iio_widget *widget,
 	widget->update_value = update_value;
 	widget->save = save;
 	widget->priv = priv;
+	widget->priv_progress = NULL;
+	widget->priv_convert_function = NULL;
```
commodo: would a `memset(widget, 0, sizeof(*widget))` work instead at the beginning of this init block ?
the added benefit is that if you extend the `iio_widget` struct later, you can be sure that the new fields are 0/NULL if un-initialized explicitly;

libini2.h
```diff
@@ -20,4 +20,6 @@ int foreach_in_ini(const char *ini_file,
 
 int ini_unroll(const char *input, const char *output);
 
+void write_driver_name_to_ini(FILE *f, const char *driver_name);
```
commodo: curios: will this be used later ?
because if it won't be, then maybe we could just leave it as static inside `libini2.c`

commodo:
Looks good.

I'd add some descriptions to the commits to detail why they're being done.
It's not quite clear what these functions will be used for, or why they are being added or exported.

Alternatively, they can be added/exported in the same commit that uses them; which means grouping these patches with commits that use each function that is added/exported.
Either way works [i.e. commit description detailing their future role [or] grouping these commits with commits that will use these functions].

libini2.h
dNechita: Yes, it will be used later. I've mentioned this in the PR description.

iio_widget.c
dNechita: Yes, that will be better.

dNechita:
I've added the requested comments.

commodo:
[LGTM](https://scontent.fdub2-2.fna.fbcdn.net/v/t1.0-9/50589056_2067228833365893_7527438199266738176_n.png?_nc_cat=107&_nc_ht=scontent.fdub2-2.fna&oh=c8fbd979dd2b5188e8161fb07f1ebe4f&oe=5CFD7462)

commodo:

PR #124, opened 2019-02-08
Multiple adrv9009 plugin no adv

dNechita:
Refactors the ADRV9009 plugin to work with multiple adrv9009-phy devices (adrv9009-phy, adrv9009-phy-b, adrv9009-phy-c, etc.)
The main logic in this refactor is to create a structure that corresponds to each adrv9009-phy device and therefore move all static variables in this structure.
Also all methods that were previously assuming only one device exists, are now refactored to handle multiple devices.
The widgets for the 2nd, 3rd, etc. device are all taken from the same .glade file but the widgets are loaded through a different gtk builder. The .glade file will continue to look the same (as it was for one device only) and tabs will be dynamically created, in the plugin, for any extra device.

The advanced plugin is not in this PR. It will be in a separate one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,7 @@ if(UNIX)
 	add_definitions(-DFRU_FILES="${CMAKE_PREFIX_PATH}/lib/fmc-tools/")
 endif()
 
-set(OSC_SRC osc.c oscplot.c datatypes.c iio_widget.c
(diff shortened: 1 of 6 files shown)


commodo:
[LGTM](https://i-h2.pinimg.com/564x/11/10/6b/11106b5fff69b6db25ccba5f99493f1f.jpg)

PR #128, opened 2019-02-11
Multiple adrv9009 advanced plugin

dNechita:


plugins/adrv9009_adv.c
```diff
@@ -731,26 +737,31 @@ static const char *adrv9009_adv_sr_attribs[] = {
...
+				plugin->handle_external_request(plugin, "Reload Settings");
 			}
 		}
 	}
 }
 
 static void signal_handler_cb(GtkWidget *widget, gpointer data)
 {
-	struct w_info *item = data;
+	struct data {
```
commodo: why not define a single struct type ?

plugins/adrv9009_adv.c
```diff
@@ -1013,107 +1030,164 @@ static int adrv9009adv_handle_driver(const char *attrib, const char *value)
...
+{
+	struct osc_plugin *plugin= g_new0(struct osc_plugin, 1);
+
+	if (!plugin_ctx ) {
+		fprintf(stderr, "Cannot create plugin because an invalid plugin context was provided\n");
+		return NULL;
+	}
+
+	plugin->priv = g_new0(struct plugin_private, 1);
+	plugin->priv->plugin_ctx.plugin_name = g_strdup(plugin_ctx->plugin_name);
```
commodo: there doesn't seem to be an equivalent free for this; or i missed it

plugins/adrv9009_adv.c
```diff
@@ -1013,107 +1030,164 @@ static int adrv9009adv_handle_driver(const char *attrib, const char *value)
...
+		char *name;
+		if (devices->len > 1)
+			name = g_strdup_printf("%s-%i", THIS_DRIVER, i);
+		else
+			name = g_strdup(THIS_DRIVER);
+
+		context->required_devices = NULL;
+		context->required_devices = g_list_append(context->required_devices, g_strdup(iio_device_get_name(dev)));
+		context->plugin_name = g_strdup(name);
+		g_free(name);
```
commodo: why not just assign `name` here and remove `g_free` ?
one less `g_strdup()`

plugins/adrv9009_adv.c
```diff
@@ -1013,107 +1030,164 @@ static int adrv9009adv_handle_driver(const char *attrib, const char *value)
...
+
+		/* Construct the name of the plugin */
+		char *name;
+		if (devices->len > 1)
+			name = g_strdup_printf("%s-%i", THIS_DRIVER, i);
+		else
+			name = g_strdup(THIS_DRIVER);
+
+		context->required_devices = NULL;
+		context->required_devices = g_list_append(context->required_devices, g_strdup(iio_device_get_name(dev)));
```
commodo: do these names get free'd as well ?

plugins/adrv9009_adv.c
dNechita: I guess I didn't find a good name for this structure so I hid it inside the two functions where it was needed. I will declare this structure only once.

plugins/adrv9009_adv.c
dNechita: There is. The osc_plugin_context_free_resources() takes care of this.

plugins/adrv9009_adv.c
dNechita: That's better, thanks!

plugins/adrv9009_adv.c
dNechita: Yes, osc_plugin_context_free_resources() takes care of these as well.

mirceacaprioru:
Why not squash the last two commits in the one that generated the discussion ? Commit 50ceb8921f157831d644faa9de4131a4b8373de0 ?

commodo:
[LGTM](http://thumbpress.com/wp-content/uploads/2013/12/Act-natural1.jpg)

PR #130, opened 2019-02-19
Adrv9009 mcs

dNechita:


plugins/adrv9009.c
```diff
@@ -261,8 +261,8 @@ static void multichip_sync()
 
 		guint n = 0;
```
commodo: why not squash this into the previous patch ?

commodo:
rest looks good

plugins/adrv9009.c
dNechita: Because it is @mirceacaprioru's work and findings.

commodo: ok
fine from me

commodo:
[LGTM](https://i-h2.pinimg.com/564x/d4/f2/00/d4f20041254a0727ddce7cb81be9e68c.jpg?b=t)

PR #131, opened 2019-02-19
plugins/adrv9009: Lower the lower bound of the AUX PLL LO freq to zero

dNechita:
The TX can be disabled and in this case the AUX PLL LO frequency will be 0. Therefore we remove the lower limit of the spin button (which was 70 Meg) otherwise the plugin when reads 0 will force the widget to 70 Meg and write back to the driver. And this is not desired.


commodo:
[LGTM](https://i-h2.pinimg.com/564x/2f/48/2e/2f482edfafccc0e6ded17c3d32ac94c9.jpg)

PR #133, opened 2019-02-22
Dac data manager fully scalable tx

dNechita:
Previous to this PR, the dac_data_manager assumed that a DDS can have a maximum of 8 tones.
Changes in this PR modify things so that the dac_data_manager can handle any number of tones.

Notes:
A group of 2 tones (I and Q) forms a channel.
A group of 2 channels (1st and 2nd) forms a TX.

So dac_data_manager used to work with a maximum of 2 TXs (TX1 and/or TX2).


plugins/dac_data_manager.c
```diff
@@ -140,7 +141,7 @@ static bool tx_channels_check_valid_setup(struct dac_buffer *dbuf);
 
 static const gdouble abs_mhz_scale = -1000000.0;
 static const gdouble khz_scale = 1000.0;
-static const char *default_channel_names[8] = {
+static const char *default_channel_names[8] = { // TO DO: Should generate these automatically if number of TX will be variable
```
commodo: this comment is just patch noise if you're (re)moving this in the next patch

plugins/dac_data_manager.c
```diff
@@ -1237,23 +1227,46 @@ static void gui_manager_create(struct dac_data_manager *manager)
...
 	/* If name convention "TX*_I|Q_F* is missing */
 	if (name && strncmp(name, "TX", 2) != 0) {
-		name = iio_channel_get_id(ch);
+		g_free(name);
+		name = g_strdup(iio_channel_get_id(ch));
 		if (name && !strncmp(name, TONE_ID, TONE_ID_SIZE)) {
 			tone_index = name[TONE_ID_SIZE];
 			if (tone_index && g_ascii_isdigit(tone_index))
-				name = default_channel_names[tone_index - '0'];
+				name = build_default_channel_name_from_index(tone_index - '0');
```
commodo: another approach, would be to use a `name_buf[32]` which we could pass to `build_default_channel_name_from_index(name_buf, tone_index - '0'`)

that way, there shouldn't be a need to do strdup()

plugins/dac_data_manager.c
```diff
@@ -141,16 +141,6 @@ static bool tx_channels_check_valid_setup(struct dac_buffer *dbuf);
 
 static const gdouble abs_mhz_scale = -1000000.0;
 static const gdouble khz_scale = 1000.0;
-static const char *default_channel_names[8] = { // TO DO: Should generate these automatically if number of TX will be variable
```
commodo: this patch adds a lot of allocation & free-ing;
which just complicates things [for both review & future bug findings];  

iio-oscilloscope wasn't particularly a stable piece of software, and this doesn't help to get close to that point

maybe, let's change things a bit; 
either:
1. we can we build/alloc this list on a single init/constructor path, and re-use the `default_channel_names` var-name ; and destroy the list of the destructor path
2.  paths can be constructed inside a buffer that is passed to the `build_` function

plugins/dac_data_manager.c
```diff
@@ -1237,23 +1227,46 @@ static void gui_manager_create(struct dac_data_manager *manager)
...
+ * success and NULL otherwise.
+ */
+static char * get_tone_name(struct iio_channel *ch)
 {
-	const char *name;
+	char *name;
 	char tone_index;
 
-	name = iio_channel_get_name(ch);
+	name = g_strdup( iio_channel_get_name(ch));
```
commodo: does this really need to be `strdup-ed` ?

plugins/dac_data_manager.c
```diff
@@ -1541,35 +1541,23 @@ static void manage_dds_mode (GtkComboBox *box, struct dds_tx *tx)
 	min_scale = manager->lowest_scale_point;
 	scale_available_mode = manager->scale_available_mode;
 
-	GtkWidget *tx1, *tx2 = NULL, *tx3 = NULL, *tx4 = NULL;
```
commodo: I like this patch :)

I can't tell if this could break things, because it seems some code paths are not common to both dac1 & dac2 ; so, this patch also may hide some weird issues.
But the overall benefit of generalizing things helps and we can fix things later when we find them

commodo: maybe make `dac1` & `dac2` into a `dac[]` array later ?
this doesn't need to be done now, but it's thought

if there is a patch that is already doing this, i'll get to it later and come back

plugins/dac_data_manager.c
```diff
@@ -1045,11 +1046,9 @@ static GtkWidget *gui_dac_create(struct dds_dac *ddac)
 
 	gtk_alignment_set_padding(GTK_ALIGNMENT(dac_align), 5, 5, 5, 5);
 
-	gtk_table_attach(GTK_TABLE(dac_table), gui_tx_create(&ddac->tx1),
```
commodo: this looks like it could be squashed into the previous patch ?

plugins/fmcomms1.c
```diff
@@ -1770,20 +1771,20 @@ static GtkWidget * fmcomms1_init(struct osc_plugin *plugin, GtkWidget *notebook,
 	if (ini_fn)
 		load_profile(NULL, ini_fn);
 
-	dds1_freq = dac_data_manager_get_widget(dac_tx_manager, TX1_T1_I, WIDGET_FREQUENCY);
-	dds2_freq = dac_data_manager_get_widget(dac_tx_manager, TX1_T2_I, WIDGET_FREQUENCY);
-	dds3_freq = dac_data_manager_get_widget(dac_tx_manager, TX1_T1_Q, WIDGET_FREQUENCY);
-	dds4_freq = dac_data_manager_get_widget(dac_tx_manager, TX1_T2_Q, WIDGET_FREQUENCY);
+	dds1_freq = dac_data_manager_get_widget(dac_tx_manager, dac_data_manager_dds_tone(0, TONE_1, TONE_I), WIDGET_FREQUENCY);
```
commodo: these also look like they could parametrized into a for loop;
i could be wrong

commodo:
this PR is a bit big; 
maybe split this into 2 ?

i admit i'm also not always doing this properly [small/manageable PRs]
typically, a PR can be reviewed well if it's ~5 changes of medium size;

this one can be split into 2 PRs of 3-4 patches each

plugins/dac_data_manager.c
```diff
@@ -2319,4 +2319,4 @@ unsigned dac_data_manager_dds_tone(unsigned tx_index,
 	enum dds_tone_index tone_index, enum dds_tone_type tone_type)
 {
 	return ((tx_index * TX_NB_TONES) + (tone_index * CHANNEL_NB_TONES) + tone_type);
-}
```
commodo: this looks like patch noise;
it could be moved into the previous patch though

plugins/debug.c
```diff
@@ -1639,8 +1639,10 @@ static GtkWidget * debug_init(struct osc_plugin *plugin, GtkWidget *notebook, co
 
 	for (i = 0; i < nb_devs; i++) {
 		struct iio_device *dev = iio_context_get_device(ctx, i);
-		gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(combobox_device_list),
-							(const gchar *)iio_device_get_name(dev));
+		const gchar *dev_name = iio_device_get_name(dev);
```
commodo: this looks a bit like a minor bug-fix;
typically the commit title should be made to highlight that it's a fix

plugins/dac_data_manager.c
```diff
@@ -1243,7 +1243,7 @@ static char *build_default_channel_name_from_index(guint ch_index)
 	guint tone_index = (ch_index % 2) + 1; /* There are always 2 tones (I/Q). Count starts at 1 */
 	const char i_q_type = (ch_index & 0x02) ? Q_CHANNEL : I_CHANNEL; /* First two indexes are I, next two are Q and so on*/
 
-	return g_strdup_printf("TX%i_%c_F%i", tx_index, i_q_type, tone_index);
+	return g_strdup_printf("TX%u_%c_F%u", tx_index, i_q_type, tone_index);
```
commodo: this looks like it's a fix commit for a change that was introduced in this PR
maybe move this change in that function
otherwise it's just noise

commodo:
> 
> 
> this PR is a bit big;
> maybe split this into 2 ?
> 
> i admit i'm also not always doing this properly [small/manageable PRs]
> typically, a PR can be reviewed well if it's ~5 changes of medium size;
> 
> this one can be split into 2 PRs of 3-4 patches each

Alright, so this PR is around 5 commits, if things get re-arranged.
We can leave it like this (and not split it now).

plugins/dac_data_manager.c
dNechita: This didn't show in the final diff.

plugins/dac_data_manager.c
dNechita: If I do that, then I need to carry a reference of the dac_data_manager all the way to this function. I don't think that it will be any better than the way it is now.

plugins/dac_data_manager.c
dNechita: I understand what you mean but I'm not using this list and there is no need to reuse the information inside that list (at the moment). The default names are used only in a particular initialization step and that is it. If this changes then yes, your suggestion should be the right approach.

plugins/dac_data_manager.c
dNechita: Yes, because name can take an allocated array as well (from build_default_channel_name_from_index() ).

plugins/dac_data_manager.c
commodo: another approach would be to build the list once on the module constructor and free it on the module destructor path;
or do, some sort of caching inside `build_default_channel_name_from_index`; 

[atm] i am not questioning whether all `strdup()`s are all free'd here; i'm questioning whether someone else who will want to extend/change this code [later], will remember to free them [given this `strdrup()` + `free()` pattern]

plugins/dac_data_manager.c
dNechita: Yes, this could be the next step to make dac_data_manager more generic.

plugins/dac_data_manager.c
dNechita: ok

plugins/fmcomms1.c
dNechita: Yes, they could be. But in a different PR.

plugins/dac_data_manager.c
commodo: following comment https://github.com/analogdevicesinc/iio-oscilloscope/pull/133/commits/d5e9b4da94f4fdbe8145685cd75beb4b6a5dabff#r259751705  , this could e converted into `static const char **default_channel_names ` and `build_default_channel_name_from_index()` constructs this list on the first call (i.e. if `default_channel_names` is NULL)

in that case all that needs to be remembered, is to de-alloc `default_channel_names` [it could be renamed to something else] when the module is destroyed or unloaded

dNechita: The list cannot be static, it needs to belong to an instance. You could have many instances of dac_data_manager (simultaneously), each with its own number of channels. E.g one instance with 2 channels and one instance with 64 channels.

PR #140, opened 2019-03-04
osc.c: Fix freeze for osc trigger

adisuciu:
Commit e0bad3c7e7d6907ee72a7173fb94cdb4d7e513a0 removed void typecasting.
From memmove() reference, the src and dest parameters need to be typecasted
to void

Signed-off-by: Adrian Suciu <adrian.suciu@analog.com>


dNechita:
Can you try to make the static analyzer happy as well?
It complains that:
> '(void*)info->data_ref' is of type 'void *'. When using void pointers in calculations, the behaviour is undefined.

I was thinking just now that it might work if you cast to void* the result of adding the two pointers. Can you try?

PR #150, opened 2019-03-29
make: Fix uninstall target

nunojsa:
The uninstall target was not removing adi-osc.desktop correctly.

Signed-off-by: Nuno Sa <Nuno.Sa@analog.com>


commodo:
[LGTM](https://i.redd.it/ewvjkm3x1po21.png)

PR #151, opened 2019-03-29
eeprom: Fix possible segfault

nunojsa:
Remove free() call which could lead to a possible segmentation fault.
The first time find_eeprom() is called, eeprom_path is NULL. Either way,
must be the caller responsibility to free the allocated memory.

Signed-ooff-by: Nuno Sa <Nuno.Sa@analog.com>

### DIFF
--- a/eeprom.c
+++ b/eeprom.c
@@ -6,45 +6,65 @@
  * Licensed under the GPL-2.
  */
 
-#include <ftw.h>
-#include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdio.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
 #include <libgen.h>
 
 #include "eeprom.h"
 
-static char *eeprom_path = NULL;
-
-/* Test if a given file is a valid XCOMM EEPROM file. */
-static int is_eeprom(const char *fpath, const struct stat *sb,
-		int typeflag, struct FTW *ftwbuf)
-{
-	int ret = 0;
-	char *fpath_s = strdup(fpath);
-	if (typeflag == FTW_F && !strcmp(basename(fpath_s), "eeprom") \
-			&& sb->st_size == FAB_SIZE_FRU_EEPROM) {
-		free(eeprom_path);
-		eeprom_path = strdup(fpath);
-		ret = 1;
-	}
-	free(fpath_s);
-	return ret;
-}
 
 /* Recursively search a given path (defaulting to /sys) for an XCOMM compatible
  * EEPROM file.
  *
  * If a matching EEPROM file is found the path is returned, otherwise returns
  * NULL. Note that the string for the returned path is obtained with malloc and
- * should be freed.
+ * should be freed. It's up to the caller to free the allocated moemroy.
  */
 const char *find_eeprom(const char *path)
 {
-	if (path == NULL)
-		path = "/sys";
+	char *eeprom_path = NULL;
+	char eeprom_names[512];
+	FILE *fp = NULL;
+	char cmd[512];
 
-	if (nftw(path, is_eeprom, 64, FTW_DEPTH | FTW_PHYS) == 1)
-		return eeprom_path;
-	return NULL;
+	if (path == NULL) {
+		path = "/sys";
+	}
+
+	snprintf(cmd, sizeof(cmd), "find %s -name eeprom 2>/dev/null", path);
+
+	fp = popen(cmd, "r");
+	if (fp == NULL) {
+		perror("popen");
+		return NULL;
+	}
+
+	while (fgets(eeprom_names, sizeof(eeprom_names), fp) != NULL) {
+		struct stat eeprom_file;
+		char *__basename = NULL;
+
+		if (eeprom_names[strlen(eeprom_names) - 1] == '\n')
+			eeprom_names[strlen(eeprom_names) - 1] = '\0';
+
+		__basename = strdup(eeprom_names);
+		stat(eeprom_names, &eeprom_file);
+
+		if (S_ISREG(eeprom_file.st_mode) &&
+		    !strcmp(basename(__basename), "eeprom") &&
+		    eeprom_file.st_size == FAB_SIZE_FRU_EEPROM) {
+
+			eeprom_path = strdup(eeprom_names);
+			free(__basename);
+			break;
+		}
+
+		free(__basename);
+	}
+
+	pclose(fp);
+	return eeprom_path;
 }


eeprom.c
```diff
@@ -24,7 +24,7 @@ static int is_eeprom(const char *fpath, const struct stat *sb,
 	char *fpath_s = strdup(fpath);
 	if (typeflag == FTW_F && !strcmp(basename(fpath_s), "eeprom") \
 			&& sb->st_size == FAB_SIZE_FRU_EEPROM) {
-		free(eeprom_path);
```
commodo: typically, `free()` handles NULL; at least for glibc;
is there a case where this isn't happening ?
i admit that there are some libc implementations that may not handle that ;

but, if you want, you could just do a NULL check;
otherwise this may risk doing leaks

nunojsa: Yes, you are right, `free()` should handle it correctly. I'm not sure if it is applicable to osc, but in multithreaded applications, that free may be an obvious problem leaving some thread pointing to memory which was freed already... Either way, typically in this API's such as `find_eeprom()` it's the caller responsibility to free the allocated memory. But, if you prefer and if the multithreaded case is not applicable to osc i can just do the null check...

commodo: ok, but if the discussion is about multithreaded stuff, then the fix is to add some sort of locking there, since it's a shared resource

or to rework it, so that it's not shared;

nunojsa: > or to rework it, so that it's not shared;
That would be the best option in my opinion. But since we are using `nftw()` might be not possible. I will check if we can rework it...

commodo: don't invest too much energy into this;
locking may be good enough as well;

nunojsa: Ok. The naming for the PR is not that correct. Indeed, there is no problem with `free(NULL)`, but assuming that the caller is freeing the memory, calling free again on already free'd memory leads to undefined behavior. For removing the shared resource, the only alternative i could come with, is doing something like [here](https://github.com/analogdevicesinc/iio-oscilloscope/blob/0573b1e6883ac21f8d7049fc3ae33312bbf13ae7/plugins/fmcomms1.c#L1403). Honestly, Im not a big fan of `popen()` and `system()` but it saves us from a shared resource. The other way would be too time consuming (using directly the API's in dirent.h). I will update the PR with the reworked function...

commodo: i don't see how double-free could happen, unless there is some weird multi-threading bug ;
`eeprom_path` is re-assigned right after being free'd;

`popen() & system()` seem a bit overkill for this ; but i wouldn't mind them much here;

using `dirent.h` shouldn't be too bad either (time consuming);
i mean, it takes 30-50 lines to write a function that could be generic;

nunojsa: > i don't see how double-free could happen

I think it would happen in the second time we make the call to `find_eeprom()`. We return the address of the allocated memory with `strdup()` to the caller which typically will call `free()` or `g_free()` ([here)](https://github.com/analogdevicesinc/iio-oscilloscope/blob/0573b1e6883ac21f8d7049fc3ae33312bbf13ae7/plugins/fmcomms2.c#L873),leaving our shared resource in `eeprom.c` pointing to an already freed region...

> but i wouldn't mind them much here;

Then if it is acceptable I will go for it since I already have a proposal...

>using dirent.h shouldn't be too bad either;

I'm just worried with this because we would need some recursion here to have the recursive search for a given directory which can be not that straight... But if you prefer this I can invest a bit more on it...

nunojsa:
I guess it also makes sense to squash these 2 commits into one?

commodo:
> I guess it also makes sense to squash these 2 commits into one?

yep :)

eeprom.c
```diff
@@ -6,45 +6,65 @@
...
+	}
+
+	while (fgets(eeprom_names, sizeof(eeprom_names), fp) != NULL) {
+		struct stat eeprom_file;
+		char *__basename = NULL;
+
+		if (eeprom_names[strlen(eeprom_names) - 1] == '\n')
+			eeprom_names[strlen(eeprom_names) - 1] = '\0';
+
+		__basename = strdup(eeprom_names);
```
commodo: `strdup()` -ing `eeprom_names` seems un-necessary if you are not modifying it & free-ing it afterwards

eeprom.c
```diff
@@ -6,45 +6,65 @@
...
+		stat(eeprom_names, &eeprom_file);
+
+		if (S_ISREG(eeprom_file.st_mode) &&
+		    !strcmp(basename(__basename), "eeprom") &&
+		    eeprom_file.st_size == FAB_SIZE_FRU_EEPROM) {
+
+			eeprom_path = strdup(eeprom_names);
```
commodo: this is still using `eeprom_path` [which is the shared resource that we should get rid of]
this isn't much better than before [when using nftw()]

commodo: sorry, i just saw that you are removing `static char *eeprom_path = NULL;`
in this case, you could just return `strdup(eeprom_names);` without modifying this code to use popen()

but you should also add in this same commit the `free(eeprom_path)` for wherever this is called;

eeprom.c
commodo: >  I think it would happen in the second time we make the call to find_eeprom(). We return the address of the allocated memory with strdup() to the caller which typically will call free() or g_free() (here),leaving our shared resource in eeprom.c pointing to an already freed region...

oh, i just saw that;
so, just remove the static pointer `static char *eeprom_path = NULL;` and make sure that it is free'd wherever `find_eeprom()` is called

eeprom.c
nunojsa: Im doing it because im calling `basename()` afterwards which will modify the passed string (in this case `eeprom_names`).

eeprom.c
nunojsa: > in this case, you could just return strdup(eeprom_names); without modifying this code to use popen()

You mean, still using `nftw()`? if so, with `nftw()` we need the global variable since all the work is done in the callback passed to `nftw()` which in our case was `is_eeprom()`. `nftw()` itself will only return an `int`.

eeprom.c
nunojsa: Yes, all the callers are doing their job :)

eeprom.c
commodo: oh right ; i forgot about that;
let's leave this as-is then

eeprom.c
commodo: ack

eeprom.c
nunojsa: :+1: . I will then squash it and update the PR

PR #152, opened 2019-04-01
[2018_R2] plugins/adrv9009: Hide MCS for single device usage and be less verbose

commodo:
In case the HMC7044 is not used, skip single shot mode workaround.
The AD9528 won't need this anyways. Hide the MCS button in single device
mode. This addresses issue #148


commodo:

PR #156, opened 2019-04-12
Fix gksu deprecated

### LINKED ISSUE #102: GKSu deprecated
The `adi-osc.desktop` file included with the installation uses `gksudo` to launch the program. Unfortunately, GKSu has been deprecated and is starting to be dropped in several distributions, such as [Ubuntu 18.04](https://itsfoss.com/gksu-replacement-ubuntu/) and [Archlinux](https://lists.archlinux.org/pipermail/arch-dev-public/2018-April/029224.html).

As a result, the desktop launcher no longer works. Would it be difficult to switch to PolicyKit?

nunojsa:
This PR aims to fix #102. It also configures and installs an adi-osc.desktop file when using cmake. Also note that, the polkit action is only being supported on cmake builds....

### DIFF
--- a/org.adi.pkexec.osc.policy.cmakein
+++ b/org.adi.pkexec.osc.policy.cmakein
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE policyconfig PUBLIC "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN" "http://www.freedesktop.org/standards/PolicyKit/1/policyconfig.dtd">
+<policyconfig>
+  <vendor>Analog Devices Inc.</vendor>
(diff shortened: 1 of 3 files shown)


org.adi.pkexec.osc.policy.cmakein
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE policyconfig PUBLIC "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN" "http://www.freedesktop.org/standards/PolicyKit/1/policyconfig.dtd">
+<policyconfig>
+  <vendor>Analog devices</vendor>
```
dNechita: Can this be "Analog **D**evices? or "Analog Devices Inc."?

nunojsa: Yes, i can make that change.

commodo:
If no other objections, will merge this later today.

PR #161, opened 2019-05-24
Spectrum Analyzer: Use integer values when working with sample rates

dNechita:
This way we avoid comparing floating point values which is not a reliable thing to do anyway.
This fixes an issue that happens on OpenSuse Leap 42.3 where such a float comparison fails (61.44 != 61.39999).

### DIFF
--- a/plugins/spectrum_analyzer.c
+++ b/plugins/spectrum_analyzer.c
@@ -68,7 +68,7 @@ typedef struct _fastlock_profile {
 
 /* Plugin Global Variables */
 static const double sweep_freq_step = 56; /* 56 MHz */
-static const double sampling_rate = 61.44; /* 61.44 MSPS */
+static const long long sampling_rate = 61440000; /* 61.44 MSPS */
 
 static struct iio_context *ctx;
 static struct iio_device *dev, *cap;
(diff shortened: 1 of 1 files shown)


plugins/spectrum_analyzer.c
```diff
@@ -159,29 +159,29 @@ static ssize_t demux_sample(const struct iio_channel *chn,
...
+		iio_channel_attr_write_longlong(ch0, "sampling_frequency", freq_hz);
 	else
 		fprintf(stderr, "Failed to retrieve iio channel in %s\n", __func__);
 }
 
-static double device_get_rx_sampling_freq(struct iio_device *dev)
+static long long device_get_rx_sampling_freq(struct iio_device *dev)
 {
-	double freq = 0.0;
+	long long freq_hz = 0.0;
```
commodo: `long long freq_hz = 0;` ?

plugins/spectrum_analyzer.c
```diff
@@ -398,11 +405,6 @@ static bool configure_data_capture(plugin_setup *setup)
 		dev_info->adc_scale = '?';
 		dev_info->adc_freq = 0.0;
```
commodo: maybe also change this to `dev_info->adc_freq = 0;`
no idea if some compiler warnings could trigger

plugins/spectrum_analyzer.c
dNechita: yes

plugins/spectrum_analyzer.c
commodo: oh; wait; `adc_freq` is still double

dNechita: `dev_info->adc_freq` remains of type float so no actions needed here

PR #165, opened 2019-07-03
plugin: cn0508: Initial revision for cn0508

mirceacaprioru:
This is an initial revision for the cn0508. This is a desktop power supply
based on the ad7124-8 adc and ad5683r dac. This plugin will enable the use
of the power supply over the network, to set different voltage and current
ranges.

### DIFF
--- a/plugins/cn0508.c
+++ b/plugins/cn0508.c
@@ -1,0 +1,318 @@
+/**
+ * Copyright (C) 2019 Analog Devices, Inc.
+ *
+ * Licensed under the GPL-2.
+ *
+ **/
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <gtk/gtk.h>
+#include <gtkdatabox.h>
+#include <glib.h>
+#include <math.h>
+#include <stdint.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <sys/stat.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "../datatypes.h"
+#include "../osc.h"
+#include "../iio_widget.h"
+#include "../osc_plugin.h"
+#include "../config.h"
+
+#define THIS_DRIVER "CN0508"
+
+#define ADC_DEVICE "ad7124-4"
+#define DAC_DEVICE "ad5683r"
+
+#define N_BITS 24
+
+#define V_REF_ADC  2.5
+#define GAIN_ADC   1
+
+struct _cn0508_data {
+	double adc_conversion;
+	double dac_output;
+	double output_current;
+	double output_voltage;
+	double u2_temp;
+	double u3_temp;
+
+	double pot_voltage;
+	double pot_current;
+
+	double input_voltage;
+};
+
+static struct _cn0508_data cn0508_data;
+
+static struct iio_context *ctx;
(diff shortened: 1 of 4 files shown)


plugins/cn0508.c
```diff
@@ -0,0 +1,323 @@
...
+}
+
+static void save_widget_value(GtkWidget *widget, struct iio_widget *iio_w)
+{
+	iio_w->save(iio_w);
+}
+
+static gboolean update_display(gpointer foo)
+{
+	int ret;
```
commodo: doesn't the compiler warn about this maybe being un-initialized?

commodo: seems it does:
https://travis-ci.org/analogdevicesinc/iio-oscilloscope/jobs/553747156

mirceacaprioru: yes it does, will fix.

plugins/cn0508.c
```diff
@@ -0,0 +1,323 @@
...
+static struct _cn0508_data cn0508_data;
+
+static struct iio_context *ctx;
+static struct iio_device *adc, *dac;
+static struct iio_channel *u2_temp_ch, *u3_temp_ch, *out_current_ch;
+static struct iio_channel *in_v_attenuator_ch, *out_v_attenuator_ch,
+	       *current_pot_pos_ch, *voltage_pot_pos_ch;
+static struct iio_channel *dac_ch;
+
+static struct iio_widget iio_widgets[100];
```
dNechita: If you think this plugin has little chance of having many widgets in the future, you can lower the number of elements to 50 or 25.

plugins/cn0508.c
```diff
@@ -0,0 +1,323 @@
...
+#include <glib.h>
+#include <gtkdatabox_grid.h>
+#include <gtkdatabox_points.h>
+#include <gtkdatabox_lines.h>
+#include <math.h>
+#include <stdint.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
dNechita: This has already been included above (2nd include)

plugins/cn0508.c
```diff
@@ -0,0 +1,323 @@
...
+ * Copyright (C) 2019 Analog Devices, Inc.
+ *
+ * Licensed under the GPL-2.
+ *
+ **/
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <gtk/gtk.h>
+#include <gtkdatabox.h>
```
dNechita: I don't think you need to include all these headers that all related to gtkdatabox.

plugins/cn0508.c
```diff
@@ -0,0 +1,323 @@
...
+#include <stdint.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "../libini2.h"
```
dNechita: If this plugin doesn't need to save/restore things, then this include is not needed.

mirceacaprioru:
Changelog V1 -> V2
- remove unnecessary includes
- initialize ret variable
- reduce the size of iio_widgets array to 25

PR #171, opened 2019-07-31
plugins/lidar: Change ADP Bias lower limit to prevent negativ raw value

dNechita:

### DIFF
--- a/glade/lidar.glade
+++ b/glade/lidar.glade
@@ -3,7 +3,7 @@
   <requires lib="gtk+" version="2.24"/>
   <!-- interface-naming-policy project-wide -->
   <object class="GtkAdjustment" id="adj_adp_bias">
-    <property name="lower">100</property>
+    <property name="lower">122</property>
     <property name="upper">200</property>
     <property name="value">160</property>
     <property name="step_increment">1</property>


glade/lidar.glade
```diff
@@ -3,7 +3,7 @@
   <requires lib="gtk+" version="2.24"/>
   <!-- interface-naming-policy project-wide -->
   <object class="GtkAdjustment" id="adj_adp_bias">
-    <property name="lower">100</property>
+    <property name="lower">122</property>
```
commodo: hmm; i am not so sure this correct; either that, or i am missing something;
there seems to be some inconsistencies on how this is handled;

the raw_convert function is:
```
static double adp_bias_volts_to_raw_convert(double value, bool inverse)          
{                                                                                
        double ret;                                                              
                                                                                 
        if (inverse) {                                                           
                ret = -((value * 5 * 18.18) / 4096) + 122;                       
        } else {                                                                 
                ret = ((-122 - value) * 4096) / (5 * 18.18);                     
        }                                                                        
                                                                                 
        return ret;                                                              
}         
```

in `set_all_iio_atributes_to_default_values()` this is called with
```
       iio_channel_attr_write_longlong(afe_ch0, "raw",                          
                adp_bias_volts_to_raw_convert(-160, false));         
```
with a negative value, so that `-122 - (-160)` works

then in `iio_widget.c` the scale (-1.0) is applied after the conversion:
```
       if (widget->priv_convert_function)                                       
                freq = ((double (*)(double, bool))widget->priv_convert_function)(freq, true);
                                                                                 
        freq /= fabs(scale);                                                     
                                                                                 
        /* if scale is negative, we treat things a little differently */         
        if (scale < 0) {                                                         
                /* if the setting is negative, and it can be set negative */     
                if (mag < 0 && min < 0)                                          
                        freq *= -1;                                              
                else if (min >= 0)                                               
                        freq *= -1;                                              
        }                                       
```

[ where, `priv_convert_function` == `adp_bias_volts_to_raw_convert` ]

in any case, the scale logic here is a bit confusing I guess

dNechita: Scale if applied first, then the conversion function. This happens when saving the widget (writing its value to the iio attribute) .
The code above is for updating the widget value, where the operations are in the reverse order: first conversion function, then scale apply.

commodo: ack;
thanks for clarification;
sorry for the noise

PR #172, opened 2019-07-31
Osc preferences

dNechita:
Add a system for the plugins to send their preferences to the osc and the osc plot.
The idea is to have a straight forward way to exchange this kind of data.
All preferences are collected by osc and merged (if no conflicts). Then the preferences are use throughout the application.
Examples of preferences:
 - default value of sample count when creating a plot window
 - constraints for some channels to be always checked in the plot window
 - constraints for some channels to be un-toggleable in the plot window
 - additional FFT correction to be used for a specific iio-device
Other preferences might come up in the future

### DIFF
--- a/osc.c
+++ b/osc.c
@@ -45,6 +45,7 @@ static int num_check_fcts = 0;
 static GSList *plugin_lib_list = NULL;
 static GSList *dplugin_list = NULL;
 static struct osc_plugin *spect_analyzer_plugin = NULL;
+static OscPreferences *osc_preferences = NULL;
 GtkWidget *notebook;
 GtkWidget *infobar;
 GtkWidget *tooltips_en;
@@ -1463,7 +1464,11 @@ GtkWidget * new_plot_cb(GtkMenuItem *item, gpointer user_data)
 {
 	GtkWidget *new_plot;
 
-	new_plot = osc_plot_new(ctx);
+	if (osc_preferences && osc_preferences->plot_preferences)
+		new_plot = osc_plot_new_with_pref(ctx, osc_preferences->plot_preferences);
+	else
+		new_plot = osc_plot_new(ctx);
+
 	osc_plot_set_visible(OSC_PLOT(new_plot), true);
 	plot_init(new_plot);
 
@@ -1587,6 +1592,11 @@ static void do_quit(bool reload)
 	}
 
 	math_expression_objects_clean();
+
+	if (osc_preferences) {
+		osc_preferences_delete(osc_preferences);
+		osc_preferences = NULL;
+	}
 }
 
 void application_reload(struct iio_context *new_ctx, bool load_profile)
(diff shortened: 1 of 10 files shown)


osc_preferences.h
```diff
@@ -0,0 +1,126 @@
...
+        chn_pref->chn_name = g_strdup(chn_name);
+        chn_pref->constraints = NULL;
+
+        return chn_pref;
+}
+
+static inline void iio_chn_preferences_free(IioChnPreferences *pref)
+{
+        if (pref->chn_name) {
+                g_free(pref->chn_name);
```
commodo: curios: is `g_free()` NULL tolerant?
so that you can skip the null check

osc.c
```diff
@@ -1463,7 +1464,11 @@ GtkWidget * new_plot_cb(GtkMenuItem *item, gpointer user_data)
 {
 	GtkWidget *new_plot;
 
-	new_plot = osc_plot_new(ctx);
+	if (osc_preferences && osc_preferences->plot_preferences)
```
commodo: the use of the global `osc_preferences` variables looks a bit weird here;
a instance is dependent on the existence of the global var;

at least in the current context, i don't see why this is needed [of it is really needed]

osc_preferences.h
```diff
@@ -0,0 +1,126 @@
...
+{
+        IioChnPreferences *chn_pref = g_new(IioChnPreferences, 1);
+
+        chn_pref->chn_name = g_strdup(chn_name);
+        chn_pref->constraints = NULL;
+
+        return chn_pref;
+}
+
+static inline void iio_chn_preferences_free(IioChnPreferences *pref)
```
commodo: just as a note: these `inline` specifiers don't do anything for these free callback/functions, since these functions are also passed via function pointers to some Glib list cleanup thingi;
inline usually works when the compiler can expand this code as a "macro"; and based on some internal logic it does or doesn't do that

they are fine to leave here though

osc_preferences.h
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/* OscPlotPreferences */
+
+static inline OscPlotPreferences* osc_plot_preferences_new()
+{
+        OscPlotPreferences *plot_pref = g_new(OscPlotPreferences, 1);
+
+        plot_pref->sample_count = NULL;
+        plot_pref->devices_pref_list = NULL;
```
commodo: just curios; does `g_new()` does [or doesn't] guarantee that the memory is zeroed-out ?

if it doesn't guarantee, one way to do this consistently is to use a `memset(plot_pref, 0)` type of construct;
this will be future-proof when you want to add new field, but forget to initialize the NULL here;

osc_preferences.h
dNechita: It is NULL tolerant. Thanks!

osc_preferences.h
dNechita: I can use g_new0() which does just that. Thanks!

osc.c
commodo: my fault here;
i misinterpreted the `osc_preferences` var-name

osc_preferences.h
dNechita: I used the inline just to avoid duplicate symbols since this is included in multiple places. I will move the functions implementation to a .c file in order to hide implementation details.

PR #173, opened 2019-08-01
oscplot: Prevent enabling/disabling the channel if it is greyed out

dNechita:


oscplot.c
```diff
@@ -6662,6 +6663,20 @@ static void capture_window_realize_cb(GtkWidget *widget, OscPlot *plot)
...
+	gboolean sensitive;
+
+	gtk_tree_model_get(model, chn_iter, SENSITIVE, &sensitive, -1);
+
+	if (sensitive) {
+		gtk_tree_store_set(GTK_TREE_STORE(model), chn_iter, CHANNEL_ACTIVE, state, -1);
+		return true;
+	}
+
+	return false;
```
commodo: this return value doesn't seem to be used;
will it be used later?

dNechita: Since this function might fail to do what is designed to do, I think is a good practice to provide a way to give feedback to the caller. This return value could be used later.

PR #180, opened 2019-09-09
build,travis-c: update build configuration

commodo:
1. Add support for Xcode 11 (implicitly also for Mac OS Mojave)
2. Reverts patch .travis.yml: force xenial for docker builds ; Xenial images should be the default now (hopefully); removing this limitation let's Travis-CI upgrade the default docker image (since Xenial is the minimum we need).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,14 +21,10 @@ matrix:
       env:
         - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
     - os: linux
-      # FIXME: remove dist, when Xenial images become default in Travis CI
-      dist: xenial
       env:
         - OS_TYPE=centos_docker
         - OS_VERSION=7
     - os: linux
-      # FIXME: remove dist, when Xenial images become default in Travis CI
-      dist: xenial
       env:
         - OS_TYPE=ubuntu_docker
         - OS_VERSION=bionic
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,11 @@ matrix:
       osx_image: xcode10.1
       env:
         - EXTRA_SSH=-oHostKeyAlgorithms=+ssh-dss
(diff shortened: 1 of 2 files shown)


commodo:
Changelog v1 -> v2:
* rebased on newer master to include fix from commit 7fa515831ed4418b2da606d4aca8ee4405492e30

rgetz:
There is an XCode 11 build failure - it would be great to include this before all the bug fixing starts - so we know what is happening....

commodo:
> There is an XCode 11 build failure - it would be great to include this before all the bug fixing starts - so we know what is happening....

well, at this point in time, i think travis-ci.org is starting to crack;
and we should probably move to travis-ci.com ;
which means switching from legacy-services integration to Github Apps;

we're slowly being kicked out; i can't seem to configure (on travis-ci.org ) things for organizations;
could just be me;

commodo:
reason for me saying the above is:
https://travis-ci.org/github/analogdevicesinc/libiio/jobs/703176772

```
  "@type": "error",
  "error_type": "insufficient_access",
  "error_message": "operation requires create_request access to repository",
  "resource_type": "repository",
  "permission": "create_request",
  "repository": {
    "@type": "repository",
    "@href": "/repo/5378690"
```

I guess it's saying i don't have access to the ADI Github group/team;
and they're not showing up in travis-ci.org
so, my Travis-CI access token can't trigger CI builds;
reminder: libiio triggers libad9361-iio which triggers iio-oscillscope and then when artifacts are publlished by the prior build, the next build runs;

i can't even seem to trigger builds manually in travis-ci.org
maybe i could by simply pushing to master, but ¯\_(ツ)_/¯

commodo:
closing this;
Travis-CI won't be the winner for the next CI;
so no need to keep this around

PR #184, opened 2019-10-01
Add inter decim

mirceacaprioru:


plugins/ad9371.c
```diff
@@ -966,6 +973,27 @@ static int compare_gain(const char *a, const char *b)
...
+	iio_channel_attr_read_double(ch, "sampling_frequency", &freq);
+
+	gtk_label_set_text(GTK_LABEL(label_sampling_freq_rx), g_strdup_printf ("%f", freq / mhz_scale));
+
+	ch = iio_device_find_channel(dds, "voltage0", true);
+	iio_channel_attr_read_double(ch, "sampling_frequency", &freq);
+	gtk_label_set_text(GTK_LABEL(label_sampling_freq_tx), g_strdup_printf ("%f", freq / mhz_scale));
+}
+
+static void int_dec_update_cb(GtkComboBox *cmb, gpointer data)
```
dNechita: This could use a better naming (there are many controls that have their values incremented/decremented). One suggestion: 'fpga_frequency_changed_cb'

plugins/ad9371.c
```diff
@@ -1656,6 +1713,7 @@ static GtkWidget * ad9371_init(struct osc_plugin *plugin, GtkWidget *notebook, c
 	/* Update all widgets with current values */
 	printf("Updating widgets...\n");
 	update_widgets();
+	sample_frequency_changed_cb(NULL);
```
dNechita: this function calls rx_freq_info_update() which is also called on the next line. So it would get called twice. Maybe you just needed the other call in rx_freq_info_update() which is glb_settings_update_labels(). So call that directly.

plugins/adrv9009.c
```diff
@@ -1224,6 +1260,34 @@ static GtkWidget *adrv9009_init(struct osc_plugin *plugin, GtkWidget *notebook,
...
+
+	ch = iio_device_find_channel(cap, "voltage0_i", false);
+	if (iio_channel_find_attr(ch, "sampling_frequency_available")) {
+		iio_combo_box_init(&fpga_widgets[num_fpga++],
+						   cap, ch, "sampling_frequency",
+						   "sampling_frequency_available",
+						   fpga_rx_frequency_available, NULL);
+	} else {
+		gtk_widget_hide(GTK_WIDGET(gtk_builder_get_object(builder,
+														  "receive_frame_dma_buf")));
```
dNechita: lots of tabs here

plugins/adrv9009.c
```diff
@@ -1351,6 +1415,9 @@ static GtkWidget *adrv9009_init(struct osc_plugin *plugin, GtkWidget *notebook,
 			iio_toggle_button_init_from_builder(&subcomponents[i].rx_widgets[subcomponents[i].num_rx++],
 							subcomponents[i].iio_dev, subcomponents[i].ch1, "powerdown", builder,
 							"rx2_powerdown_en", 0);
+
```
dNechita: All signals connection are done a little bit latter. Look in source code for line: /* Connect signals */. Is there any reason this was needed here?

plugins/ad9371.c
mirceacaprioru: the name comes from interpolation decimation -> int_dec it was a copy paste from pluto but will go with the naming suggetion.

plugins/adrv9009.c
mirceacaprioru: the only reason to call it here is to tie it to label_sampling_frequency that corresponds to each device.

mirceacaprioru:
Changelog V1->V2
- freed strings returned by g_strdup_vprintf
- moved signals connection to /* Connect signals */ and removed the Widget object associated with the callback
- added iio_widgets_update to update_witdgets function
- removed sample_frequency_changed_cb call from init function

PR #196, opened 2019-10-29
Handle deprecated gtk symbols

dNechita:


dialogs.c
```diff
@@ -914,18 +914,20 @@ gint create_blocking_popup(GtkMessageType type, GtkButtonsType button,
...
 	glong ret = 0;
 
-	parsed = g_time_val_from_iso8601(iso8601_date, &time);
-	if (parsed) {
-		ret = time.tv_sec - build_time;
-	} else {
+	GDateTime *time = g_date_time_new_from_iso8601(iso8601_date, NULL);
+	if (time == NULL) {
 		printf("%s could not parse date. Not a ISO 8601 format.", __func__);
+		return -1;
```
commodo: returning -1 here will lead to a dialog saying `This software is newer than the latest release`
maybe the codepaths could be fixed where this is called; or just return 0 ?

Makefile
```diff
@@ -65,6 +65,7 @@ CFLAGS := $(DEP_CFLAGS) \
 	-DFRU_FILES=\"$(FRU_FILES)\" $(GIT_VERSION) \
 	$(GIT_COMMIT_TIMESTAMP) \
 	-DOSC_VERSION=\"$(OSC_VERSION)\" \
+	-DGTK_DISABLE_DEPRECATED \
```
commodo: do we need a similar one in CMake ?

dialogs.c
```diff
@@ -925,9 +927,22 @@ static glong date_compare_against_build_date(const char *iso8601_date)
...
+#else
+
+	GTimeVal time;
+	gboolean parsed;
+
+	parsed = g_time_val_from_iso8601(iso8601_date, &time);
+	if (parsed) {
+		ret = time.tv_sec - build_time;
+	} else {
+		printf("%s could not parse date. Not a ISO 8601 format.", __func__);
```
commodo: this one doesn't return anything here; as is above;
which is a inconsistent behavior between GLib versions;

commodo:
generally, some commit descriptions would be nice;

commodo:
only 2 more important notes; rest looks good

dNechita:
Changelog v0 -> v1:
 - Fix comparing date by not using the current date (which I added by mistake). 
 - Returning 0 is better than -1 in case the comparison cannot be made.
 - Add the GTK_DISABLE_DEPRECATED definition to cmake
 - Say more things for some commits

dNechita:
Thanks @commodo for looking this careful at things when doing the review!

PR #197, opened 2019-11-08
Fix comboboxtext issue

dNechita:
Recently, we've stopped using some deprecated methods of GtkComboBox and started using the recommended methods that belong to class GtkComboBoxText.
However, GtkComboBoxText objects that are loaded from glade files do not function properly.

GtkComboBoxText has a bug that makes it unsuable (can't get the active text, or insert new text, etc.). In order to work, instances of GtkComboBoxText need the property "entry-text-column" >= 0, which doesn't happen when a GtkComboBoxText is obtained from a glade file (creating them with new works).

To fix this, there's now a script which can parse glade files and add the missing property.

An additional script has been added which should be used to fix glade files edited with Glade versions that target GTK3 and not GTK2.

### DIFF
--- a/NOTE_FOR_DEVELOPERS.txt
+++ b/NOTE_FOR_DEVELOPERS.txt
@@ -1,0 +1,1 @@
+1. Every time a glade is edited with Glade editor for GTK3 (versions 3.9, 3.10, 3.11, etc.) the fixup scripts from dev-tools need to be run for the modified glade file. However if Glade 3.8 is used the fixup scripts are no longer required.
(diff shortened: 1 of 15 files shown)


NOTE_FOR_DEVELOPERS.txt
```diff
@@ -0,0 +1 @@
+1. Every time a glade is edited with Glade editor for GTK3 (versions 3.9, 3.10, 3.11, etc.) the fixup scripts from dev-tools need to be run for the modified glade file. However if Glade 3.8 is used the fixup scripts are no longer required.
```
commodo: maybe add something to Travis-CI run that tries to do the fixup, and if someone forgets, fail the build and cat this text-file/message to make sure things don't slip through

commodo:
just one comment

rest looks good

NOTE_FOR_DEVELOPERS.txt
dNechita: @commodo I've added a pre build script that does just that. Let me know if it's ok with you.

PR #201, opened 2019-11-13
ci,os-x: reduce build times by not upgrading

commodo:
Travis-CI OS X images have plenty of stuff installed. And they are updated
every now-n-then. Running `brew upgrade` can take too much on older images.

So, don't upgrade packages. Just use the ones installed.


commodo:
seems the build is not happy with bison

commodo:
Closing for a moment.
Seems OS X build fails with some new failure. Will fix

PR #207, opened 2019-11-28
Plot chn sort

dNechita:


dNechita:
Things don't work as expected at run-time. I'm currently testing things and will come back with some updates. Until then, reviewing this PR can be skipped.

iio_utils.c
```diff
@@ -1,7 +1,21 @@
...
+
 #include <string.h>
+#include <ctype.h>
+
+static gint iio_chn_cmp_by_name(gconstpointer ptr_a, gconstpointer ptr_b)
+{
+	struct iio_channel *ch_a = *(struct iio_channel **)ptr_a;
+	struct iio_channel *ch_b = *(struct iio_channel **)ptr_b;
+
+	const char *name_a = iio_channel_get_name(ch_a) ?: iio_channel_get_id(ch_a);
```
commodo: nitpick: indentations is inconsistent;

iio_utils.c
```diff
@@ -1,7 +1,21 @@
...
+        const char *name_b = iio_channel_get_name(ch_b) ?: iio_channel_get_id(ch_b);
+
+        return str_natural_cmp(name_a, name_b);
```
commodo: can `name_a` or `name_b` be NULL?
`str_natural_cmp` does not seem to handle NULL; should it?

iio_utils.c
```diff
@@ -19,3 +33,62 @@ GArray * get_iio_devices_starting_with(struct iio_context *ctx, const char *sequ
...
+ * rather than comparing the individual digits.
+ *
+ * This makes sure that e.g. in_voltage9 is placed before in_voltage10
+ */
+int str_natural_cmp(const char *s1, const char *s2)
+{
+        unsigned int n1, n2;
+	unsigned int i1 = 0, i2 = 0;
+
+        while (s1[i1] && s2[i2]) {
```
commodo: nitpick: indentation is inconsistent

oscplot.c
```diff
@@ -3766,8 +3766,11 @@ static void device_list_treeview_init(OscPlot *plot)
 		plot_channels_add_device(plot, dev_name);
 		priv->nb_input_devices++;
 
-		for (j = 0; j < nb_channels; j++) {
-			struct iio_channel *ch = iio_device_get_channel(dev, j);
+		GArray *channels = get_iio_channels_naturally_sorted(dev);
+
+		for (j = 0; j < channels->len; ++j) {
+			struct iio_channel *ch = g_array_index(channels,
+				struct iio_channel *, j);
```
commodo: this needs to update an error path, otherwise it leaks;
```
			pic = plot_iio_channel_new(priv->ctx);
			if (!pic)
				return;
```
should be   `break` instead of `return`

dNechita:
Update regarding the run-time issues: there are no issues.
This PR has been created due to the fact that the iio channels of a iio device were not provided in the usual order which is alphanumerical sorted. As it turns out this happened only on one product and it was caused by some bug in the Linux driver which is being fixed.
**Important**: now the question is do we proceed with this PR?
One one hand iio-osc will not depend anymore on the order of iio channels returned by libiio. On the other hand the performance will have a minor impact since after channels are read from device they get sorted.

commodo:
> the performance will have a minor impact since after channels are read from device they get sorted.

i think the performance impact is negligible against the added benefit of sorting the channels.
It's only done once on start AFAICT.

iio_utils.c
dNechita: name_a and name_b can't be NULL - I will document this
str_natural_cmp should not handle NULL

commodo: > str_natural_cmp should not handle NULL

`str_natural_cmp()` maybe shouldn't
but what about `iio_chn_cmp_by_name()` ?

dNechita: Which pointer did you have in mind? The function parameters (ptr_a, ptr_b) or the data that they point to? (ch_a, ch_b)

dNechita:
> > the performance will have a minor impact since after channels are read from device they get sorted.
> 
> i think the performance impact is negligible against the added benefit of sorting the channels.
> It's only done once on start AFAICT.

Yes, the sort happens once per plot construction. Then each time the current device in the  'saveas dialog' is changed. And last, when user enables/disables a plot channel in frequency mode. None of these happen frequently.

oscplot.c
dNechita: good catch

iio_utils.c
commodo: so, if you would do:
```
if (name_a == name_b)
   return 0;
```
this handles the NULL case for both;

and if you do:
```
if (!name_a)
     return -strlen(name_b);

if (!name_b)
     return strlen(name_a);
```

then there is something that works for the other NULL cases;
i am not sure where this looks best; `iio_chn_cmp_by_name()` or `str_natural_cmp()`; wherever this is put, it does look like a good idea to handle NULL

dNechita: This suggestion could be done and work but what it also does is hide programming errors.
str_natural_cmp() compares strings as the name suggests. Comparing a string with NULL doesn't make sense.
iio_chn_cmp_by_name() is designed to work on a collection (GArray). If by some reason the array doesn't contain valid data, iio_chn_cmp_by_name() should not crash and signal a warning which currently does. But shouldn't handle somebody else's error.

commodo: Fine by me

PR #221, opened 2020-02-28
Use glade3 8 oscplot

dNechita:


commodo:
[maybe a discussion]
maybe we are fighting a bit too much with GTK2;
what would it take to convert it to GTK3?
the reason for asking is EZ support;

dNechita:
> [maybe a discussion]
> maybe we are fighting a bit too much with GTK2;
> what would it take to convert it to GTK3?
> the reason for asking is EZ support;
 The plotting widget (gtkdatabox) works much slower when with GTK3. At least that is what I remember. This is the only reason we continued with GTK2.

dNechita:
> the reason for asking is EZ support;

What do you mean?

nunojsa:
> This is the only reason we continued with GTK2.

Maybe using GTK3 in some embedded boards is also to heavy? Just guessing :smile:

mhennerich:
GTK3 databox uses some different graphics library (was it cairo?), which we found damn slow without HW acceleration. That's why we continued to use GTK2.
Other applications such as the JESD GTK Eyscan already uses GTK3. It's only the plotting part which concerned us. Maybe we should re-benchmark GTK3 vs. GTK2 on Zynq!

commodo:
> Maybe we should re-benchmark GTK3 vs. GTK2 on Zynq!

I also wanted to take the discussion in that direction a bit.
But I focused on other stuff and forgot about it.

For now we can have this as-is, but it does feel like we're burying ourselves into more work and EZ-support if we do more of these types of changes.

mhennerich:
If we come to the conclusion that latest GTK3 runs with acceptable speed, we should definitely update.
I'm less concerned about  EZ-support, more on maintenance. Right now we use glade3 to design GUIs, which isn't really compatible with GTK2. So we use some search and replace scripts on the XML files, to just make them work again. However this is breaking Glade somehow, and causes all kind of weirdness when you later edit those files.

commodo:
We can try this on the latest ADI Kuiper Linux.

dNechita:
> Other applications such as the JESD GTK Eyscan already uses GTK3. It's only the plotting part which concerned us. Maybe we should re-benchmark GTK3 vs. GTK2 on Zynq!

I think this should be the next step, yes. So I'm pausing this PR for now.

commodo:
> > Other applications such as the JESD GTK Eyscan already uses GTK3. It's only the plotting part which concerned us. Maybe we should re-benchmark GTK3 vs. GTK2 on Zynq!
> 
> I think this should be the next step, yes. So I'm pausing this PR for now.

so, the main problem now is gtkdatabox;
it's a bit stuck at using GTK2;
that would mean we'd need to carry this over ourselves as a custom-build over GTK3;

PR #222, opened 2020-03-11
plugins/generic_plugin: Initial version for a generic plugin

cristiantpop:
The generic plugin is used when no other plugin has been created and
provides a means of loading a data file and sending it as a buffer to a dac
device. It creates a gui panel for each device with dac properties.

Signed-off-by: Cristian Pop <cristian.pop@analog.com>

### DIFF
--- a/glade/generic_plugin.glade
+++ b/glade/generic_plugin.glade
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <requires lib="gtk+" version="2.24"/>
+  <!-- interface-naming-policy project-wide -->
+  <object class="GtkVBox" id="generic_panel">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <child>
+      <object class="GtkNotebook" id="plugin_notebook">
+        <property name="visible">True</property>
+        <property name="can_focus">True</property>
+        <child>
+          <object class="GtkAlignment" id="alignment1">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <child>
+              <object class="GtkVBox" id="dds_vbox">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <child>
+                  <placeholder/>
+                </child>
+                <child>
+                  <placeholder/>
+                </child>
+              </object>
+            </child>
+          </object>
+        </child>
+        <child type="tab">
+          <object class="GtkLabel" id="label1">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="yalign">0.51999998092651367</property>
+            <property name="label" translatable="yes">Controls</property>
+          </object>
+          <packing>
+            <property name="tab_fill">False</property>
+          </packing>
+        </child>
+        <child>
+          <placeholder/>
+        </child>
+        <child type="tab">
+          <placeholder/>
+        </child>
+        <child>
+          <placeholder/>
+        </child>
+        <child type="tab">
+          <placeholder/>
+        </child>
+      </object>
+      <packing>
+        <property name="expand">True</property>
+        <property name="fill">True</property>
+        <property name="position">0</property>
+      </packing>
+    </child>
+  </object>
+</interface>
--- a/osc.c
+++ b/osc.c
@@ -928,6 +928,29 @@ static void load_plugin_finish(GtkNotebook *notebook,
 	plugin_make_detachable(d_plugin);
 }
 
+static void generic_plugin_handle(struct osc_plugin *generic_plugin, void *generic_lib, const char *ini_fn) {
+	struct params {
+		struct osc_plugin *plugin;
+		GtkWidget *notebook;
+		const char *ini_fn;
+	} *params;
+
+	if (!generic_plugin || !generic_lib)
+		return;
+
+	plugin_list = g_slist_append (plugin_list, (gpointer) generic_plugin);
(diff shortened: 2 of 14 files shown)


plugins/generic_plugin.c
```diff
@@ -0,0 +1,206 @@
...
+
+	for (i = 0; i < dev_count; i++) {
+		dac = iio_context_get_device(ctx, i);
+		dac_tx_manager = dac_data_manager_new(dac, NULL, ctx);
+		if (!dac_tx_manager)
+			continue;
+
+		crt_dev_name = iio_device_get_name(dac);
+		if (g_slist_find_custom(dac_dev_names, crt_dev_name, compare_func))
+			continue;
```
dNechita: There's a memory leak happening here: at this point there is memory allocated for dac_tx_manager and not freed anywhere.

plugins/generic_plugin.c
```diff
@@ -0,0 +1,206 @@
...
+
+		can_update_widgets = true;
+	}
+
+	if (!generic_en) {
+		osc_destroy_context(ctx);
+		return NULL;
+	}
+
+	return generic_panel;
```
dNechita: dac_dev_names needs to be freed before leaving

plugins/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@
 #  Boston, MA 02110-1301, USA.
 
 set(PLUGINS
+	generic_plugin
```
dNechita: Please add the plugin to Makefile as well. We're still using it.

plugins/generic_plugin.c
```diff
@@ -0,0 +1,206 @@
...
+
+static bool can_update_widgets;
+
+static void tx_update_values(void)
+{
+	iio_update_widgets(tx_widgets, num_tx);
+}
+
+static int compare_gain(const char *a, const char *b) __attribute__((unused));
+static int compare_gain(const char *a, const char *b)
```
dNechita: compare_gain is not used in this plugin

plugins/generic_plugin.c
```diff
@@ -0,0 +1,206 @@
...
+		if (g_slist_find_custom(dac_dev_names, crt_dev_name, compare_func))
+			continue;
+
+		generic_en = true;
+
+		dds_container = gtk_frame_new("DDS");
+		gtk_container_add(GTK_CONTAINER(dds_vbox), dds_container);
+		gtk_container_add(GTK_CONTAINER(dds_container),
+				  dac_data_manager_get_gui_container(dac_tx_manager));
+		gtk_widget_show_all(dds_container);
```
dNechita: After this you need to set the range of the frequency widget that belongs to the dac_data_manager. By default the range is +- 100Hz which is a very narrow range. The widget won't be able to update with the hardware value.
So you need to get the sampling frequency of the DAC, divide it by 2 and then you get range needed for the dac_data_manager. Since this plugin is generic, getting the sampling frequency might not be possible since the steps to do that can differ from DAC to DAC.
So what you can do is to set the range to be very large:
For e.g.
dac_data_manager_freq_widgets_range_update(dac_tx_manager, INT_MAX / 2);

glade/generic_plugin.glade
```diff
@@ -0,0 +1,192 @@
...
+          </object>
+          <packing>
+            <property name="position">1</property>
+          </packing>
+        </child>
+        <child type="tab">
+          <object class="GtkLabel" id="label2">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="label" translatable="yes">Block Diagram</property>
```
dNechita: Is the Block Diagram tab really necessary for this generic plugin?

plugins/generic_plugin.c
cristiantpop: Fix memory leak.

plugins/generic_plugin.c
cristiantpop: Fixed memory leak.

plugins/CMakeLists.txt
cristiantpop: Done.

plugins/generic_plugin.c
cristiantpop: Removed unused code.

plugins/generic_plugin.c
cristiantpop: Set the range of the frequency widget that belongs to the dac_data_manager.

glade/generic_plugin.glade
cristiantpop: Remove block diagram from generic plugin.

osc.c
```diff
@@ -928,6 +928,28 @@ static void load_plugin_finish(GtkNotebook *notebook,
...
+	plugin_list = g_slist_append (plugin_list, (gpointer) generic_plugin);
+	plugin_lib_list = g_slist_append(plugin_lib_list, generic_lib);
+
+	params = malloc(sizeof(*params));
+	params->plugin = generic_plugin;
+	params->notebook = notebook;
+	params->ini_fn = ini_fn;
+
+	GtkWidget *widget = init_plugin(params);
+	load_plugin_finish(GTK_NOTEBOOK(notebook), widget, generic_plugin);
```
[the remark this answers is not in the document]
cristiantpop: Check the return value of "init_plugin" function.

cristiantpop:
V2:

- Fix memory leak;

- Add the plugin to Makefile;

- Removed unused code, related to "compare_gain";

- Set the range of the frequency widget that belongs to the dac_data_manager;

- Remove block diagram from "generic_plugin.glade";

- Check the return value of "init_plugin" function.

PR #229, opened 2020-03-25
plugins: Rename generic_plugin to generic_dac

dbogdan:
Since this plugin was created for managing the DAC devices that don't have
dedicated plugins, use a more appropriate name.

Signed-off-by: Dragos Bogdan <dragos.bogdan@analog.com>


dNechita:
That's a really good point!
There still are some references to the generic plugin in osc.c. Could you please rename them as well?

dbogdan:
Sure, I'll do that.

dbogdan:
v2:
 - Rename osc.c references.

PR #233, opened 2020-03-26
AD9081 plugin dev

mhennerich:
Let's start review and merge this. There is still some functionally that needs to be added, but we'll do this later on.

### DIFF
--- a/plugins/ad9081.c
+++ b/plugins/ad9081.c
@@ -32,6 +32,7 @@
 #define NUM_MAX_WIDGETS			81
 
 const gdouble mhz_scale = 1000000.0;
+const gdouble k_scale = 1000.0;
 
 struct plugin_private {
 	/* plugin context */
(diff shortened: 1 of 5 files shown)


commodo:
rebase on newer master?

dbogdan:
It should be rebased and get_dac_dev_names() function should be implemented too. I can do that.

mhennerich:
Sounds good thanks.

dbogdan:
I've only done the rebase. I will create a different pull-request for the other changes I'm proposing for the ad9081-plugin-dev branch.

commodo:
there are still merge commits in this PR

mhennerich:
rebased it on origin/master

plugins/ad9081.c
```diff
@@ -0,0 +1,618 @@
...
+/*
+ * 81 = 2 widgets per maximmum of 8 rx channels + 8 widgets per maximun 8 tx
+ * channels + 1 combo box for global RX configuration...
+ */
+#define NUM_MAX_WIDGETS			162
+
+const gdouble mhz_scale = 1000000.0;
+const gdouble k_scale = 1000.0;
+
+const char *dac_name;
```
dNechita: I'm not sure that dac_name should be set global. Is this plugin going to be instanced more than once? If so, all instances will return the same pointer through call get_dac_dev_names().
I know this was implemented in a separated PR but I missed this issue.

mhennerich: The plugin is going to be instantiated multiple times. So this must not be global.
@dbogdan added this only very recently.

Practically only the instance who has a RX buffer will also have the DAC buffer.
So for the Quad MxFE use case this shouldn't matter. But if you have multiple independent instances this will lead to an issue.

dbogdan: I'll move the dac_name to plugin_private and I'll pass osc_plugin to get_dac_dev_names(). I'll create a patch right now.

dbogdan: See: https://github.com/analogdevicesinc/iio-oscilloscope/pull/236 and https://github.com/analogdevicesinc/iio-oscilloscope/pull/237.

commodo:
i've rebased on the newer master;
let's see how this builds

PR #234, opened 2020-03-26
plugins: ad9081: Implement get_dac_dev_names()

dbogdan:
For preventing the generic_dac plugin to be loaded (since AD9081 has a
dedicated plugin), get_dac_dev_names() should be implemented.

Signed-off-by: Dragos Bogdan <dragos.bogdan@analog.com>

### DIFF
--- a/plugins/ad9081.c
+++ b/plugins/ad9081.c
@@ -34,6 +34,8 @@
 const gdouble mhz_scale = 1000000.0;
 const gdouble k_scale = 1000.0;
 
+const char *dac_name;
+
 struct plugin_private {
 	/* plugin context */
 	struct osc_plugin_context plugin_ctx;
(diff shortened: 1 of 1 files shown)


dbogdan:
My supposition here was that the DAC device that has scan_elements (has buffer) is called "axi-ad9081-tx-hpc".

dbogdan:
Let me use get_iio_devices_starting_with() instead... similar to ad9081_init().

dbogdan:
v2:
 - make use of get_iio_devices_starting_with() called by ad9081_init().

dNechita:
> v2:
> 
>     * make use of get_iio_devices_starting_with() called by ad9081_init().

I can't see where get_iio_devices_starting_with() is used.

dbogdan:
It is used earlier: https://github.com/analogdevicesinc/iio-oscilloscope/blob/ad9081_get_dac_dev_names/plugins/ad9081.c#L431

PR #242, opened 2020-06-22
plugins: Add support for adrv9002

nunojsa:
This adds a plugin for the adrv9002 platform. It assumes more than one
device can exist so that the plugin is dynamically created

### DIFF
--- a/glade/adrv9002.glade
+++ b/glade/adrv9002.glade
@@ -1,0 +1,2739 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.2 -->
+<interface>
+  <object class="GtkAdjustment" id="adjustment_atten_tx1">
+    <property name="lower">-41.950000000000003</property>
+    <property name="step_increment">0.050000000000000003</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_atten_tx2">
+    <property name="lower">-41.950000000000003</property>
+    <property name="step_increment">0.050000000000000003</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_gain_rx1">
+    <property name="upper">36</property>
+    <property name="step_increment">0.5</property>
+    <property name="page_increment">5</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_gain_rx2">
+    <property name="upper">36</property>
+    <property name="step_increment">0.5</property>
+    <property name="page_increment">5</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_lo_rx1">
+    <property name="lower">30</property>
+    <property name="upper">6000</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_lo_rx2">
+    <property name="lower">30</property>
+    <property name="upper">6000</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_lo_tx1">
+    <property name="lower">30</property>
+    <property name="upper">6000</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_lo_tx2">
+    <property name="lower">30</property>
+    <property name="upper">6000</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_nco_rx1">
+    <property name="lower">-20000</property>
+    <property name="upper">20000</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment_nco_rx2">
+    <property name="lower">-20000</property>
+    <property name="upper">20000</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkVBox" id="adrv9002_panel">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="border_width">5</property>
+    <property name="spacing">5</property>
+    <child>
+      <object class="GtkNotebook" id="plugin_notebook">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <child>
+          <object class="GtkVBox" id="box12">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <child>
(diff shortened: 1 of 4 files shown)


glade/adrv9002.glade
```diff
@@ -0,0 +1,2723 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.2 -->
```
dNechita, line 2: Why hasn't this been created with Glade 3.8?

plugins/adrv9002.c
```diff
@@ -0,0 +1,1241 @@
...
+#include "../iio_utils.h"
+#include "../config.h"
+#include "dac_data_manager.h"
+
+#define THIS_DRIVER "ADRV9002"
+#define PHY_DEVICE "adrv9002-phy"
+#define DDS_DEVICE "axi-adrv9002-tx"
+#define CAP_DEVICE "axi-adrv9002-rx"
+
+#define ADRV002_NUM_CHANNELS	2
```
dNechita, line 27: '9' is missing (ADRV9002)

plugins/adrv9002.c
```diff
@@ -0,0 +1,1241 @@
...
+#define NUM_MAX_WIDGETS	10
+#define NUM_MAX_DDS	2
+#define NUM_MAX_ADC	2
+
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
+
+const gdouble mhz_scale = 1000000.0;
+const gdouble k_scale = 1000.0;
+
+struct adrv9002_gtklable {
```
dNechita, line 39: label (typo)

glade/adrv9002.glade
nunojsa: It's just what you can get with ubuntu 20.04. I even tried to used docker to use older glade for gtk2 but it was not working that good at all. So,  It was easier just using this and run the fixup scripts...

plugins/adrv9002.c
nunojsa: ack

plugins/adrv9002.c
nunojsa: ack

glade/adrv9002.glade
dNechita: If Glade for GTK2 is unreliable/unusable then we should talk about this and decide whether continue using it (and telling people to use it) or not.

Did you also run this fixup script? (https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/dev-tools/fixup-comboboxtext.pl)

nunojsa: > If Glade for GTK2 is unreliable/unusable

I mean, with docker it was not working that well. And for newer distributions it's just hard to actually find a glade 3.8 package. Even, with docker I actually had to go to ubuntu 14.04 for this...

I just used `fixup-gtk3-glade.sh`

dNechita: We're building it from source (https://ftp.gnome.org/pub/GNOME/sources/glade3/3.8/). It compiles on 18.04.4 but I haven't tried on newer versions.

plugins/adrv9002.c
```diff
@@ -0,0 +1,1241 @@
...
+	combo_box_manual_update(combo);
+}
+
+static void save_gain_ctl(GtkWidget *widget, struct adrv9002_common *chann)
+{
+	char *gain_ctl;
+
+	combo_box_save(widget, &chann->gain_ctrl);
+
+	gain_ctl = gtk_combo_box_text_get_active_text(
```
dNechita, line 210: gain_ctl needs to be freed

glade/adrv9002.glade
nunojsa: Ahh, I see. Didn't knew that. Not sure how it will work on ubuntu 20.04. Do you want me to try it? I can check it but I guess if it starts to become time consuming we can just use this glade file (since we are on the run)

plugins/adrv9002.c
nunojsa: Ohh then I need to fix this in some other places. I just copied it from another plugin where I guess we have the same problem.

glade/adrv9002.glade
dNechita: It think we should use this glade file as it is. Please run the fixup-comboboxtext.pl script on it.

plugins/adrv9002.c
dNechita: I'll write this down to make sure we will go over all plugins and free where it is needed.

glade/adrv9002.glade
nunojsa: Ok. In the next time, I will try to remember this. At my office machine I run arch linux which does have glade-gtk2 build that works. But on ubuntu there's none and I tried to do it with docker and sharing the X env but that worked horribly....

plugins/adrv9002.c
nunojsa: ack. In adrv9009 we definitely have the same issue

nunojsa:
Changes in v2:

1. Fixed some typos;
2. Free strings returned by `gtk_combo_box_text_get_active_text`;
3. Run `fixup-comboboxtext.pl` on the glade file.

plugins/adrv9002.c
```diff
@@ -0,0 +1,1255 @@
...
+
+	f = fopen(file_name, "r");
+	if (!f)
+		goto err;
+
+	fseek(f, 0, SEEK_END);
+	size = ftell(f);
+	rewind(f);
+	buf = malloc(size);
+	if (!buf)
```
dNechita, line 560: 'f' should get close before jumping to 'err'

plugins/adrv9002.c
```diff
@@ -0,0 +1,1255 @@
...
+		context->required_devices = g_list_append(context->required_devices,
+							  g_strdup(iio_device_get_name(dev)));
+		context->plugin_name = name;
+		g_array_append_val(data, context);
+	}
+
+	g_array_free(devices, FALSE);
+
+	return data;
+}
```
dNechita, line 1255: an empty line is missing

plugins/adrv9002.c
```diff
@@ -0,0 +1,1255 @@
...
+
+/* Max nr of widgets per channel */
+#define NUM_MAX_WIDGETS	10
+#define NUM_MAX_DDS	2
+#define NUM_MAX_ADC	2
+
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
+
+const gdouble mhz_scale = 1000000.0;
+const gdouble k_scale = 1000.0;
```
dNechita, line 37: Looks like this is not used in this plugin. Could be removed.

plugins/adrv9002.c
nunojsa: definitely

plugins/adrv9002.c
nunojsa: Yeps. It was in the beginning but now it's just a leftover...

plugins/adrv9002.c
nunojsa: ack.

Issue #245, opened 2020-06-22
osc over Ethernet is not stable...


rgetz:
periodically, the capture will freeze.

In the console, we see...

```
FPS: 9
FPS: 8
FPS: 8
FPS: 10
FPS: 7
FPS: 8
Error while reading data: (32) Broken pipe
```

clicking on stop/start will re-start things.

rgetz:
I tried with something like this:

```
diff --git a/osc.c b/osc.c
index d07a465..844b43c 100644
--- a/osc.c
+++ b/osc.c
@@ -1241,8 +1241,13 @@ static gboolean capture_process(void *data)
                while (true) {
                        ssize_t ret = iio_buffer_refill(dev_info->buffer);
                        if (ret < 0) {
-                               fprintf(stderr, "Error while reading data: %s\n", strerror(-ret));
+                               fprintf(stderr, "Error while reading data: (%zi) %s\n", -ret, strerror(-ret));
                                stop_sampling();
+                               if (ret == -EPIPE ) {
+                                       stop_capture = FALSE;
+                                       capture_setup();
+                                       break;
+                               }
                                goto capture_stop_check;
                        }
 
```

and that mostly fixes things - but it still fails sometimes...

rgetz:
closed with #248

PR #248, opened 2020-07-07
osc: Try to recapture samples when encountering error: broken pipe

dNechita:
Addresses #245


rgetz:
this seem to work for me.

thanks

PR #253, opened 2020-07-16
Adrv9001 misc

nunojsa:
This PR adds support for TX NCO support and some minor enhancements...

### DIFF
--- a/plugins/adrv9002.c
+++ b/plugins/adrv9002.c
@@ -149,6 +149,22 @@ static void save_widget_value(GtkWidget *widget, struct iio_widget *iio_w)
 	iio_w->update(iio_w);
 }
 
+static void save_gain_value(GtkWidget *widget, struct adrv9002_common *chann)
+{
+	char *gain_ctl = gtk_combo_box_text_get_active_text(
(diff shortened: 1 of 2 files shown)


plugins/adrv9002.c
```diff
@@ -514,6 +520,38 @@ static void adrv9002_check_channel_status(struct plugin_private *priv,
...
+					    const bool tx)
+{
+	int ret;
+	long long dummy;
+	char gtk_str[32], label_str[32];
+	struct iio_widget *nco;
+
+	if (tx) {
+		nco = &priv->tx_widgets[channel].nco_freq;
+		sprintf(gtk_str, "nco_freq_rx%d", channel + 1);
```
dNechita, line 549: Not sure, but should string "nco_freq_rx" have tx instead of rx?

nunojsa: Yeps. It works because on the other branch I'm actually doing tx :facepalm: . Will do a v2...

nunojsa:
Changes in v2:

- Format the rx/tx widget and label strings in correct `if() ... else` branch.

PR #254, opened 2020-07-21
iio_utils: sort multiple devices

mhennerich:
In case multiple devices are present we assume them sorted like this:

adrv9009-phy-d
adrv9009-phy-c
adrv9009-phy-b
adrv9009-phy-a

or

adrv9009-phy-d
adrv9009-phy-c
adrv9009-phy-b
adrv9009-phy

or

adrv9009-phy-3
adrv9009-phy-2
adrv9009-phy-1
adrv9009-phy-0

adrv9009-phy-3
adrv9009-phy-2
adrv9009-phy-1
adrv9009-phy

### DIFF
--- a/iio_utils.c
+++ b/iio_utils.c
@@ -20,6 +20,23 @@ static gint iio_chn_cmp_by_name(gconstpointer ptr_a, gconstpointer ptr_b)
 	return str_natural_cmp(name_a, name_b);
 }
 
+static gint iio_dev_cmp_by_name(gconstpointer ptr_a, gconstpointer ptr_b)
+{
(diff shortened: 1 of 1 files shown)


dNechita:
Is there an advantage to sort the devices descending?

mhennerich:
Yes - I have done this on purpose.
With the new JESD-FSM framework, when updating the transceiver profile (profile_config), we need to do this on all devices,
however the link control is done only at the master device. (typically the device without suffix). So this needs to be configured last. I'll document this behavior (naming convention) on the ADRV9009 plugin description.

PR #255, opened 2020-07-22
Connect dialog updates

rgetz:
this will update the connect dialog box, so handles discovery easier for people, putting local, usb and IP contexts in a pull down, but allowing by hand manual, or serial context. Let me know if there are any issues.

image

### DIFF
--- a/dialogs.c
+++ b/dialogs.c
@@ -45,6 +45,7 @@ struct _Dialogs
 	GtkWidget *connect;
 	GtkWidget *connect_fru;
 	GtkWidget *connect_iio;
+	GtkWidget *connect_attrs;
 	GtkWidget *ctx_info;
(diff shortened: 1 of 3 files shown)


dialogs.c
```diff
@@ -1196,15 +1409,15 @@ void dialogs_init(GtkBuilder *builder)
 		if (ctx) {
 			iio_context_destroy(ctx);
 		} else {
-			GtkWidget *local = GTK_WIDGET(gtk_builder_get_object(
+/*			GtkWidget *local = GTK_WIDGET(gtk_builder_get_object(
```
dNechita, line 1412: It this commented block is not needed anymore, it should be removed.

rgetz: fixed.

PR #256, opened 2020-07-23
plugins: adrv9009: fix crash with adrv9008-1

commodo:
The current code crashes with ADRV9008-1, since not all IIO devices are
present on that board. This crashes on adrv9009_init().

Added null-checks.


commodo:
Changelog v1 -> v2:
* added `plugins: adrv9009: fix crash with adrv9008-1 (part 2)`

PR #265, opened 2020-08-24
oscplot.c: move the fps display to the GUI

rgetz:
When running on the GUI, you can't see the fps calculation, so display
it in the window.

Also remove it from printing on the console, since that is annoying.

Signed-off-by: Robin Getz <robin.getz@analog.com>


dNechita (approved):
Looks good

PR #267, opened 2020-08-28
plugins: ad9371: check iio contexts during init

Liambeguin:
Make sure the IIO contexts are valid before continuing.
In some cases different devicetree node names prevent the application
from starting because of a segmentation fault. fix it.

Signed-off-by: Liam Beguin <lvb@xiphos.com>


dNechita (approved):
Looks good. Thanks!

PR #273, opened 2020-10-09
[RFC] plugin cleanup on init failure

nunojsa:
The motivation for this was a crash provoked by a failure in initializing the adrv9002 plugin. If the plugins fails to initialize, it will do some cleanups and since this plugins provides `get_dac_dev_names()`, the generic_dac plugin will call it and thus, will try to access nonexistent data.

The reason why this is a RFC is because it's not clear to me if it is intentional that a plugin is kept around even if it fails to initialize. The natural thing to me is that if a plugin fails to initialize, then it should be removed from our plugin list and it's resources should also be cleaned....

With this change, it's clear that a plugin should not clear any data during the init callback (it might still choose to do so if it coordinates things with the `destroy()` callback). Hence, some plugins might need to be fixed to take this into account. In this series, there are already fixes for two plugins...

### DIFF
--- a/osc.c
+++ b/osc.c
@@ -832,6 +832,18 @@ static void close_plugins(const char *ini_fn)
 	plugin_lib_list = NULL;
 }
 
+static void close_plugin(struct osc_plugin *plugin, const char *ini_fn, const bool do_destroy)
+{
+	if (plugin->destroy && do_destroy)
+		plugin->destroy(plugin, ini_fn);
+
+	plugin_lib_list = g_slist_remove(plugin_lib_list, plugin->handle);
+	dlclose(plugin->handle);
+	plugin_list = g_slist_remove(plugin_list, plugin);
+	if (plugin->dynamically_created)
+		g_free(plugin);
+}
+
 static struct osc_plugin * get_plugin_from_name(const char *name)
 {
 	GSList *node;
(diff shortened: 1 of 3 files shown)


commodo (approved):

dNechita:
I don't know any reason for why we keep the plugin in list in case it fails to initialize. Maybe we had the happy flow in mind, where init() always succeeds (probably because we first call identify() which checks if the right iio devices are there). So I think this is a good fix.
Do you plan to add fixes for other plugins?

dNechita (approved):

nunojsa:
> Do you plan to add fixes for other plugins?

Yes, I can sweep other plugins to check if they might suffer from this problem. I added the fix for these two plugins since I was involved in their development and already knew they had this issue.

If you prefer I can send a v2 of this PR with more possible fixes for other plugins. Otherwise, I can just open a different PR for those

dNechita:
> > Do you plan to add fixes for other plugins?
> 
> Yes, I can sweep other plugins to check if they might suffer from this problem. I added the fix for these two plugins since I was involved in their development and already knew they had this issue.
> 
> If you prefer I can send a v2 of this PR with more possible fixes for other plugins. Otherwise, I can just open a different PR for those

A second PR sounds good. This PR can be merged.

PR #277, opened 2020-10-21
Fix installation paths and location of glade files / icons

spacefrogg:
Several paths were hard-coded, which defied cmake's ability to work with
arbitrary installation paths.

Fixes a bug which defines GLADE_FILES_DEST but does not use it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,7 +186,7 @@ set(CMAKE_PREFIX_PATH ${CMAKE_SYSTEM_PREFIX_PATH})
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 	configure_file(org.adi.pkexec.osc.policy.cmakein ${CMAKE_CURRENT_BINARY_DIR}/org.adi.pkexec.osc.policy @ONLY)
 	install(FILES ${CMAKE_CURRENT_BINARY_DIR}/org.adi.pkexec.osc.policy
-		DESTINATION /usr/share/polkit-1/actions/)
+		DESTINATION ${CMAKE_INSTALL_PREFIX}/share/polkit-1/actions/)
 	# setup exec script
 	configure_file(osc-wrapper.cmakein ${CMAKE_CURRENT_SOURCE_DIR}/osc-wrapper @ONLY)
 	install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/osc-wrapper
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -186,7 +186,7 @@ set(CMAKE_PREFIX_PATH ${CMAKE_SYSTEM_PREFIX_PATH})
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 	configure_file(org.adi.pkexec.osc.policy.cmakein ${CMAKE_CURRENT_BINARY_DIR}/org.adi.pkexec.osc.policy @ONLY)
 	install(FILES ${CMAKE_CURRENT_BINARY_DIR}/org.adi.pkexec.osc.policy
-		DESTINATION /usr/share/polkit-1/actions/)
+		DESTINATION ${CMAKE_INSTALL_PREFIX}/share/polkit-1/actions/)
```
nunojsa, line 189: Did you actually tested this? I'm not really sure, but I think polkit is looking actions in this location, so I'm not sure this will really work...

CMakeLists.txt
```diff
@@ -199,31 +199,30 @@ set(SHARE_DEST share/osc)
...
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 	configure_file(adi-osc.desktop.cmakein ${CMAKE_CURRENT_BINARY_DIR}/adi-osc.desktop @ONLY)
 	install(FILES ${CMAKE_CURRENT_BINARY_DIR}/adi-osc.desktop
-		DESTINATION /usr/share/applications/)
+		DESTINATION ${CMAKE_INSTALL_PREFIX}/share/applications/)
 
 	# install theme icons
 	install(FILES icons/osc16.png RENAME adi-osc.png
-		DESTINATION /usr/share/icons/hicolor/16x16/apps)
+		DESTINATION ${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/16x16/apps)
```
nunojsa, line 215: Ditto... Did you also tested this? I believe icon files are really intended to be in the right places...

nunojsa (commented):
You also need to provide a `Signed-off-by` tag on your commit message...

CMakeLists.txt
spacefrogg: Well, if you choose a different `PREFIX` from `/usr` this will install polkit-1 actions under the new prefix. This is, how `PREFIX` usually works. Otherwise distributions have no means to divert the output to another location for packaging. Don't the CI tests check this? I tested it on a distribution that needs to package the program into a different location from `/usr`. There, it works as intended.

CMakeLists.txt
spacefrogg: Ditto.

CMakeLists.txt
nunojsa: Yeah, I know it will install in another location. My point is, will it work?? I'm not so sure it will since polkit is looking for actions in `/usr/share/polkit-1/actions/` (AFAIK)? i don't think CI will test this. To test this, you need to see if you can run osc with sudo rights.

nunojsa: > on a distribution that needs to package the program into a different location from /usr

So this looks like a valid thing to me for this change... Please add your `Signed-of` tag to the commit...

spacefrogg:
Signed off.

nunojsa (approved):

CMakeLists.txt
mirceacaprioru: polkit does not work anymore. This change is not ok.

spacefrogg:
What distribution did you use? What were your install commands?

mirceacaprioru:
It is debian based the problem is that polkit always searches in /usr/share/polkit-1/actions/  for actions. This is not dependent on IIO-scope install path it is a different application.

spacefrogg:
What commands, exactly, did you use to install it?

mirceacaprioru:
```
cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DCMAKE_COLOR_MAKEFILE=OFF -Bbuild -H.
```
This command does not work because there are still leftovers that point to /usr/local/
https://github.com/analogdevicesinc/iio-oscilloscope/search?q=%2Fusr%2Flocal

```
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_COLOR_MAKEFILE=OFF -Bbuild -H.
```
This works but action for policy gets installed under /usr/local because of this 
https://github.com/analogdevicesinc/iio-oscilloscope/blob/92cd68c682156dcd5ad26e923754be88ffa34ba3/CMakeLists.txt#L189

nunojsa:
> This command does not work because there are still leftovers that point to /usr/local/

I wonder why this command is not working for polkit... Using `-DCMAKE_INSTALL_PREFIX=/usr` should install the polkit actions in the right place... Also, a lot of `/usr/local` leftovers look to be related with plain make build. We should definitely deprecate that and just use cmake. Having both does not make sense... For the profiles things, I guess we should have them as `.cmakein` files and let cmake generate the .ini files dependent on `DCMAKE_INSTALL_PREFIX`...

> This works but action for policy gets installed under /usr/local because of this

On debian, running osc with root fails without polkit? I'm curious since in arch and ubuntu mate 20.04 looks like they did something and I do not really need the polkit action anymore...

mirceacaprioru:
> I wonder why this command is not working for polkit... Using -DCMAKE_INSTALL_PREFIX=/usr should install the polkit actions in the right place... Also, a lot of /usr/local leftovers look to be related with plain make build. We should definitely deprecate that and just use cmake. Having both does not make sense... For the profiles things, I guess we should have them as .cmakein files and let cmake generate the .ini files dependent on DCMAKE_INSTALL_PREFIX...
I agree with this it is something that we need to get sorted.
>On debian, running osc with root fails without polkit? I'm curious since in arch and ubuntu mate 20.04 looks like they did something and I do not really need the polkit action anymore...
I am using Debian Buster and it seems i still need it.

spacefrogg:
You're right. I'll fix that and make a new PR.

PR #279, opened 2020-10-27
Plugin cf axi tdd

nunojsa:
This PR adds some additions before actually adding the cf_axi_plugin driver:

* It makes some functions that are being widely used statically in plugins generic (to avoid more code duplication);
* It adds a new combo box API without refreshing every time the combo box available options so that we can do `widget->save() followed by widget->update()` without falling in an infinite loop;
* Blocks the gtk signals during widget->update().

The last point, which is the third patch is kind of a RFC patch. It's, arguably, a "nitpick" improvement. The point, is that when we update a widget, we can receive a different value from the iio dev from the one in the GUI (eg: a failed call to widget->save() or an autonomous update). In this case, we don't really want our widget signal handler to be called on a value that we we know the device is already holding. Hence, we prevent an unneeded call to the signal handler. However, the implementation is not as straight as I thought it would be in the beginning, so if we feel that this is to much for the behavior/performance improvement we might get (since this is only a "problem" in rare situations), we might just drop this (I'm presenting it here since I had the work :smile: ). I'm also implementing all of this with new API's which also feels a bit like duplication. Hence, if we think that this change is worthy, I guess we could just implement it in the generic `iio_widget_update()` and plugins that don't actually set `sig_handler_data` won't use this anyways...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ PLUGINS=\
 	plugins/lidar.$(SO) \
 	plugins/dmm.$(SO) \
 	plugins/debug.$(SO) \
+	plugins/cf_axi_tdd.$(SO) \
 	$(if $(WITH_MINGW),,plugins/spectrum_analyzer.so) \
 	$(if $(WITH_MINGW),,plugins/scpi.so)
 
--- a/glade/cf_axi_tdd.glade
+++ b/glade/cf_axi_tdd.glade
@@ -1,0 +1,1545 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <requires lib="gtk+" version="2.24"/>
+  <!-- interface-naming-policy project-wide -->
+  <object class="GtkAdjustment" id="adjustment1">
+    <property name="upper">10000</property>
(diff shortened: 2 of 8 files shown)


nunojsa:
Changes in v2:
1. Replace spaces with tabs.

nunojsa:
Changes in v3:
1. Replace `ENONET` by `ENOENT`.

dNechita (approved):
This looks fine to me

nunojsa:
@dNechita, what about patch  3 "iio_widget: Block gtk signals on widget update"? Do you prefer to have a separate API (as is right now) or is it ok to just move the signal block functionality into the current, existing API as I suggest in the description?

dNechita:
> @dNechita, what about patch 3 "iio_widget: Block gtk signals on widget update"? Do you prefer to have a separate API (as is right now) or is it ok to just move the signal block functionality into the current, existing API as I suggest in the description?

Yes, I think it would be best to keep it as a separate API. I know this duplicates some API calls but on the other hand it is more clear what each function does.

PR #281, opened 2020-11-11
Fix installation paths

spacefrogg:
polkit needs its own prefix:
- FHS distributions have polkit actions in /usr regardless of installation path
- non-FHS distributions don't and need this configurable

Convert profiles/*.ini files to *.cmakein when they reference other files and
need to honor CMAKE_INSTALL_PREFIX.

Fixes problems with #277

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,7 +237,17 @@ endif()
 
 install(DIRECTORY icons DESTINATION ${SHARE_DEST})
 
-foreach(plib_dest xmls filters waveforms profiles block_diagrams)
+file(GLOB PROFILES profiles/*.cmakein)
+foreach(PROFILE IN LISTS PROFILES)
+  string(REPLACE ".cmakein" "" PROFILE_TARGET ${PROFILE})
+  configure_file(${PROFILE} ${CMAKE_CURRENT_BINARY_DIR}/profiles/${PROFILE_TARGET})
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/profiles/${PROFILE_TARGET} DESTINATION ${PLIB_DEST}/profiles/)
+endforeach()
+
+file(GLOB PROFILES profiles/*.ini)
+install(FILES ${PROFILES} DESTINATION ${PLIB_DEST}/profiles/)
+
+foreach(plib_dest xmls filters waveforms block_diagrams)
 	install(DIRECTORY ${plib_dest} DESTINATION ${PLIB_DEST})
 endforeach()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,8 @@ if (WIN32)
 	string(TIMESTAMP BUILD_YEAR "%Y")
 endif()
 
+include(GNUInstallDirs)
+
 # Get the GIT hash of the latest commit
 include(FindGit OPTIONAL)
 if (GIT_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
@@ -162,6 +164,10 @@ if(CMAKE_COMPILER_IS_GNUCC)
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -Wall -Wextra -Wno-unused-parameter")
 endif()
 
+# Override on distributions that do not use a global polkit prefix (e.g. NixOS)
+if(NOT DEFINED CMAKE_POLKIT_PREFIX)
+  set(CMAKE_POLKIT_PREFIX /usr)
+endif()
 
 add_library(osc ${OSC_SRC})
 target_link_libraries(osc ${LIBOSC_LIBS})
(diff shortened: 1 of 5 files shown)


marv:
I would suggest using GNUInstallDirs and `CMAKE_INSTALL_DATADIR`[1] for this (and preferly other provided `CMKAE_INSTALL_` variables where suitable)

On Exherbo[2] we use an arch-specific prefix (e.g. `/usr/x86_64-pc-linux-gnu` or `/usr/armv7-unknown-linux-gnueabihf`) to allow multiple architectures side-by-side and set `-DCMAKE_INSTALL_DATAROOTDIR:PATH=/usr/share/` in order to have all arch-agnostic files installed to `/usr/share`[3].

[1] https://cmake.org/cmake/help/v3.19/module/GNUInstallDirs.html
[2] https://www.exherbo.org/
[3] https://git.exherbo.org/arbor.git/tree/exlibs/cmake.exlib#n185

nunojsa:
In general I'm ok with @marv suggestions, however:

> ...to allow multiple architectures side-by-side and set...

Not sure if we really want to have multiple architectures installed side-by-side... We want just want to run on the specific arch we are installing... On the other hand, using GNUInstallDirs looks like a nice addition since we are just manual setting the installation  directories and GNUInstallDirs might handle some things automatically for us (depending on distribution for instance)...

@spacefrogg , I would split this into 2 commits. One for the polkit stuff and the other for the cmakein files...

spacefrogg:
Poke.

commodo:
> Poke.

sorry about taking longer here
we hit the Travis-CI-not-being-free-anymore issue;
and we are switching our CI to Azure Pipelines (we'll see about getting some sort of paid subscription to avoid getting hit again like we were with Travis-CI)

right now, we are migrating libiio to Azure, libad9361-iio is half-way there, and then next will be this repo;
we also want to switch Appveyor to Azure Pipelines;
we've been using it for Windows builds

nunojsa:
I see that `CMAKE_POLKIT_PREFIX` should be overridden for non FHS distros as they do not use polkit... Is this right @spacefrogg? I wonder if we do not need to also take action in [osc-wrapper](https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/osc-wrapper.cmakein)? Is `pkexec` available at all in NixOs for example? Or do we still have polktit and it's just that we need to install the actions somewhere other that `/usr/share/pokit/actions`?

spacefrogg:
Exactly! non-FHS distros just set `CMAKE_POLKIT_PREFIX` to deviate.

Polkit is available and patched up in a way to not only look at `/usr/share/polkit/actions`. osc-wrapper works the same way.

CMakeLists.txt
```diff
@@ -208,27 +212,27 @@ install(FILES ${GLADE_FILES} ${ICON_FILES} DESTINATION ${GLADE_FILES_DEST})
...
 	install(FILES icons/osc.svg RENAME adi-osc.svg
-		DESTINATION ${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/apps)
+		DESTINATION ${CMAKE_INSTALL_DATADIR}/icons/hicolor/scalable/apps)
 
 	# This is a bit "ugly". It makes sure that the hicolor theme directory updates
 	# it's modification date. This makes sure that osc icons are loaded right away.
 	# We can also just remove this and assume the user has do this manually or run
 	# something like `gtk-update-icon-cache -f /usr/share/icons/hicolor`.
-	install(CODE "EXECUTE_PROCESS(COMMAND sh -c \"touch /usr/share/icons/hicolor\")")
+	install(CODE "EXECUTE_PROCESS(COMMAND sh -c \"touch ${CMAKE_INSTALL_DATADIR}/icons/hicolor\")")
```
nunojsa, line 235: I think here we need to use `CMAKE_INSTALL_FULL_DATADIR`...

nunojsa (commented):
Just a minor issue and then it´s good to me...

CMakeLists.txt
spacefrogg: No, that's fine as it is. Using `CMAKE_INSTALL_FULL_DATADIR` would prevent you to install to the prefix `/usr/local`. The `CMAKE_INSTALL_FULL_*` variants are exactly for the case of polkit, where you want to ignore the install prefix and always install to a fixed location.

spacefrogg: Unless, of course, icons are of the same category as polkit...

spacefrogg: Looking at https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html, the XDG specifies that it is searched for under `/usr` and `/usr/local`. So, what to do when somebody uses `/opt` or something completely different as install prefix?

I would like to keep it as it is right now. Now, the icons are installed under the prefix, making them readily accessible for `/usr` and `/usr/local` and need further administrative care for all other prefixes, which is likely the expected behavior. It especially avoids writing to `/usr` when installing to `/opt` and the likes.

nunojsa: > No, that's fine as it is. Using CMAKE_INSTALL_FULL_DATADIR would prevent you to install to the prefix /usr/local.

Don't think so. Did you try to run a make install on a normal distro? It will try to use `share/icons/hicolor` to run the touch command which won't exist... When you run `install` with `DESTINATION`, things work because (as specified in cmake [install](https://cmake.org/cmake/help/latest/command/install.html)):

"...If a relative path is given it is interpreted relative to the value of the CMAKE_INSTALL_PREFIX variable..."

> The CMAKE_INSTALL_FULL_* variants are exactly for the case of polkit, where you want to ignore the install prefix 

AFAIU, the FULL_DIR stuff actually works by, typically, prepending the  install prefix... Check [here](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html):

"If the value is not already an absolute path, an absolute path is constructed typically by prepending the value of the CMAKE_INSTALL_PREFIX variable."

> So, what to do when somebody uses /opt or something completely different as install prefix?

I guess in this case, there's no miracles and icons won't work :smile:... But I would say, that's for the user to care.

spacefrogg: You're right. I don't know why I was sure that it worked for me before.

dNechita (approved):

nunojsa (approved):
Nice works! Thanks!

Issue #190, opened 2019-10-09
Switch completely to CMake


dNechita:
Having both a Makefile and CMake can be confusing for both users and developers. Also it doubles the maintenance effort.
CMake is a modern tool for handling build processes.

"We need to switch completely to the CMake approach, but need to make sure all other shared libraries get removed." (See #185)

nunojsa:
Done with #282

PR #282, opened 2020-11-12
Makefile: drop file; should be unused

commodo:
cmake building has been supported for about 2 releases now.
Remove the old Makefile.
We shouldn't support it now.


dNechita:
Windows build still uses Makefile. We need to update the windows script also.

dNechita:
This is where make file is used: https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/CI/appveyor/build.sh.inc#L32

nunojsa:
Finally this is being addressed :+1: . Just to add to what Dan already said:

[config.h](https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/config.h) should also be removed as part as this commit or PR. It's only here because it's needed for make. On cmake builds, it will be generated from [config.h.cmakein](https://github.com/analogdevicesinc/iio-oscilloscope/blob/master/config.h.cmakein). Not sure if there are similar cases for other files...

nunojsa (approved):

nunojsa:
Updated this PR to finally remove Make. My changes are:

* Updated gitignore;
* Removed unneeded stuff from config.h.cmakein;
* Improved the auto generated profiles handling;
* Some cosmetics related with spaces vs tabs

We still need for #349 to get merged before merging this one. But I needed to push these changes already today to this branch. If all goes well, we just need to rebase this branch once the other PR gets merged and we are good to go.

This PR fixes #190

nunojsa:
Just rebased on top of master... We should be ready to finally get rid of the legacy Makefile :)

.gitignore
```diff
@@ -8,4 +8,17 @@ core
 *~
 *.swp
 build
+CMakeCache.txt
```
dNechita, line 11: What is the preferred method? In source build or separate build dir?

dNechita (approved):

.gitignore
nunojsa: Personally, I prefer to have a different build directory. Typically just ${CMAKE_SOURCE_DIR}/build for me

dNechita: Ok. I only asked because if we strongly wanted to use a different build directory then we shouldn't add build generated files in the ignore list in order to not encourage that practice.

PR #290, opened 2020-12-01
Plugins init cleanup2

nunojsa:
This PR is a follow up on #276 . Since the changes are pretty simple and easy to review, I'm adding all the patches in one PR. I want to enforce that I'm only fixing the most obvious leaks and not going into too much details on the plugins init function.


dNechita (approved):
Looks good to me

PR #291, opened 2020-12-02
plugins: adrv9002: Support stream loading

nunojsa:
Add support to load a new stream binary in the plugin. It must be done
before loading the correspondent profile.

### DIFF
--- a/plugins/adrv9002.c
+++ b/plugins/adrv9002.c
@@ -111,6 +111,7 @@ struct plugin_private {
 	gint this_page;
 	gint refresh_timeout;
 	char last_profile[PATH_MAX];
+	char last_stream[PATH_MAX];
 	struct adrv9002_gtklabel temperature;
 	/* rx */
 	struct adrv9002_rx rx_widgets[ADRV9002_NUM_CHANNELS];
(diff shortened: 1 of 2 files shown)


plugins/adrv9002.c
```diff
@@ -603,30 +604,80 @@ static void check_nco_support(struct plugin_private *priv)
...
+	if (priv->last_stream[0])
+		gtk_file_chooser_set_filename(chooser, priv->last_stream);
+	else
+		gtk_file_chooser_set_filename(chooser, "(None)");
+}
+
+static void load_profile(GtkFileChooser *chooser, gpointer data)
+{
+	struct plugin_private *priv = data;
+	char *file_name = gtk_file_chooser_get_filename(chooser);
```
commodo, line 672: is this missing a `g_free(file_name);`?

nunojsa: It should be freed... Is it not? The only change I've make to this function was to use the new helper. But it might a bug lying there from the beginning...

commodo: oh; it is;
apologies;
the patch is a bit wonky too look at

commodo (approved):

dNechita (approved):

PR #294, opened 2020-12-16
Merge master 2019 r2

mhennerich:
There are some fixes which should be in 2019_R2 wondering if there is anything which must not be.
@nunojsa is the navassa stream loading support causing issues in case the driver doesn't (yet) support it?


nunojsa:
> @nunojsa is the navassa stream loading support causing issues in case the driver doesn't (yet) support it?

Don't think so. If not in the driver, writing the attr will fail and we should just see a pop up message in the plugin...

nunojsa (approved):

Issue #303, opened 2021-01-15
Plot update


cuciureansergiu:
After changing the sampling frequency for an ADC, the FFT plot is showing the data based on the old sampling rate.
For example,  on a sinewave of 10khz:
default samping rate, reads 10khz
1/2 sampling rate, reads 20khz,
1/4 sampling rate, reads 40khz
Each time the sampling frequency is changed, osc has to be restarted (closing the plot window and opening another one won't fix this).

mhennerich:
Where are you seeing this?
I think this is typically handled by the plugins.

cuciureansergiu:
Here are some screenshots for a 5 khz sinewave and changing only the sampling frequency for the axi_adc:

![image](https://user-images.githubusercontent.com/16073026/104729161-b3bd8e80-5740-11eb-930d-946feffe3435.png)
![image](https://user-images.githubusercontent.com/16073026/104729200-c5069b00-5740-11eb-8897-4948852d2d4e.png)

After i restart osc

![image](https://user-images.githubusercontent.com/16073026/104729347-f97a5700-5740-11eb-8cb3-1e86f93dc914.png)

Seems like the graph data is stored once, when osc starts

mhennerich:
Typically the plugin calls rx_update_device_sampling_freq(). In your case you use the DEBUG plugin and manipulate attributes.

dNechita:
One idea is to have oscplot guess which attribute is for changing the sampling frequency (if plugin doesn't call rx_update_device_sampling_freq() or if there is no plugin). Then oscplot would read that attribute exactly when the start button gets clicked. This way, the sampling rate would get updated next time the capture is started which is better than having to restart the entire aplication.

dbogdan:
Can't we check the sampling frequency each time the START button is pressed?

dbogdan:
@dNechita , I was thinking the same. :)

mhennerich:
I guess we do this already for computing the timeout in capture_setup()

dNechita:
A fix for this has been made. Closing this issue.

PR #306, opened 2021-01-20
osc: Update sampling frequency of ADC when starting a capture

dNechita:
Plugins signal the oscplot as soon as the sampling frequency is changed (usually the control for the sampling frequency is available in the plugin).
However some ADCs might now have an associated plugin but sampling frequency could be changed by using the 'Debug' plugin or even outside of the application. For this scenario, we can't know instantly that the sampling frequency has changed but an improvement is to read the sampling frequency attribute of the ADC right before starting a capture.

This addresses #303

### DIFF
--- a/osc.c
+++ b/osc.c
@@ -1482,6 +1482,8 @@ static int capture_setup(void)
 			if (timeout > min_timeout)
 				min_timeout = timeout;
 		}
+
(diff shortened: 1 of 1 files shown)


commodo (approved):

mhennerich (approved):
Let's see if we run into issues with that - maybe we should do this only when no plugin is connected. We have enough time between now and the next release...

PR #313, opened 2021-03-05
CI/appveyor: Get the compiled dependencies from a Github release

dNechita:
The artifacts stored in the appveyor build expire after 6 months. Since
the iio-osc-mingw gets rarely modified we're better off downloading the
dependecies from the artifacs section of the latest iio-osc-mingw
release.


nunojsa (approved):
Looks good. But since you are here, I would really like if we could just move the windows build to cmake? Then, I could follow on #282

dNechita:
I've created #314 to track this.

PR #319, opened 2021-04-01
plugins: dac_data_manager: fix I and Q frequency mismatch

nunojsa:
The goal of this change is to handle the frequency attr change for the I type channels. In single/dual tone modes, the Q channel is not visible to the user and the plugin automatically changes the gtk widget causing its signal to be called and thus, changing the value in the device. The problem is that for some frequency values, we won't get exactly what we tried to set (due to integer approximations in the driver) which means that 'widget->update(widget)' called from 'iio_widget_save()' will trigger
another gtk signal to set the frequency on the Q channel for the value we got from the driver. Thus, we will end up with slightly different setting in the I and Q channels. With this change, we make use of the '*_block_signals_by_data' to save the widget value and make sure that 'widget->update(widget)' won't trigger another signal in case we get a different value.

This was noticed when testing low rate profiles in the adrv9002 plugin. In that case, the mismatch between I and Q was really causing a very poor spectrum (not that noticeable in high rate profiles). It feels that this is not a big of an issue for the scale and phase attributes but maybe @dNechita  or @mhennerich might have a better overview regarding that...

### DIFF
--- a/plugins/dac_data_manager.c
+++ b/plugins/dac_data_manager.c
@@ -1460,6 +1460,34 @@ static void dds_locked_freq_cb(GtkToggleButton *btn, struct dds_tx *tx)
 	dds_locked_phase_cb(NULL, tx);
 }
 
+/*
+ * The goal of this API is to handle the frequency attr change for the I type channels.
+ * In single/dual tone modes, the Q channel is not visible to the user and the plugin automatically
+ * changes the gtk widget causing its signal to be called and thus, changing the value in the device.
+ * The problem is that for some frequency values, we won't get exactly what we tried to set
+ * (due to integer approximations in the driver) which means that 'widget->update(widget)'
+ * called from 'iio_widget_save()' will trigger another gtk signal to set the frequency on the Q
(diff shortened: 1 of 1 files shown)


dNechita:
I think this is ok. But I'm not sure if it is needed for scale and phase.

mhennerich:
In case there is the chance that scale on I vs Q is different or that the phase between I & Q is not 90 deg.
We need to fix it. Any variation will cause bad Quadrature error.

nunojsa:
> In case there is the chance that scale on I vs Q is different or that the phase between I & Q is not 90 deg.
We need to fix it. Any variation will cause bad Quadrature error.

Ok, I did a quick check and for the scale we should be fine as we don't update the widget after saving the value. However, for the phase, if there's a chance that the driver will set a value which is not exactly what we tried to configure, we will have the same situation. I will have to do some testing for the phase attr...

nunojsa:
Tested the phase attr and it works as-is...

dNechita (approved):

PR #323, opened 2021-04-22
Fix axi core tdd

nunojsa:
Fix TDD plugin which was not loading. With this patch we are now using IIO labels to match the devices...


nunojsa:
The error in appveyor:

https://ci.appveyor.com/project/analogdevicesinc/iio-oscilloscope/builds/38828908#L1343

can be discarded. It just means that we need to update the CI job to use an up to date libiio (master one) that supports labels...

dNechita (approved):

PR #326, opened 2021-04-28
plugins: adrv9002: improve carrier handling

nunojsa:
Handling the carrier frequency is not really straight because it looks
like we have 4 independent controls when in reality they are not, since
the device only has 2 LOs. It all the depends on the LO mappings present in
the current profile. First, the only way to guarantee that a carrier is actually applied (PLL
re-tunes) is if all ports on the same LO, are moved into the calibrated
state before changing it. Not doing so means that we may end up in an
inconsistent state. For instance, consider the following steps in a FDD
profile where RX1=RX2=LO1 (assuming both ports start at 2.4GHz):

 1. Move RX1 carrier to 2.45GHz -> LO1 re-tunes
 2. Move RX1 back to 2.4GHz -> LO1 does not re-tune and our carrier is at
    2.45GHz

With the above steps we are left with both ports, __apparently__, at
2.45GHz but in __reality__ our carrier is at 2.4GHz.

Secondly, when both ports of the same type are at the same LO, which
happens on FDD and TTD (with diversity) profiles, we should move the
carrier of these ports together, because it's just not possible to have
both ports enabled with different carriers (they are on the same LO!). On
TDD profiles, we never move TX/RX ports together even being on the same LO.
The assumption is that we might have time to re-tune between RX and TX
frames. If we don't, we need to manually set TX and RX carriers to the same
value before starting operating...

### DIFF
--- a/plugins/adrv9002.c
+++ b/plugins/adrv9002.c
@@ -36,6 +36,7 @@
 #define NUM_MAX_ADC	2
 
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
+#define ADRV9002_LO_INV	255
 
 const gdouble mhz_scale = 1000000.0;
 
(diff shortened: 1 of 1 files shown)


dNechita (commented):
I would avoid using two underscore at the beginning of the variable name. As far as I know they are reserved for the compiler. This could create weird bugs.

nunojsa:
> I would avoid using two underscore at the beginning of the variable name. As far as I know they are reserved for the compiler. This could create weird bugs.

Arghh, forgot about this. I think  int the struct definition would be safe, but for consistency I will change it everywhere...

nunojsa:
Changes in v3:

* Removed leading double underscores from variable names.

dNechita (approved):

PR #329, opened 2021-05-14
plugins: adrv9002: improve ORx handling

nunojsa:
With the latest changes in the adrv9002 driver, if ORx is enabled, we won't
be able to change any control that might trigger an ensm change in
the RX/TX of the same channel as the enabled ORx. Hence, we control the
widgets sensitivity accordingly...

Special care is needed for tracking calibrations. The thing with tracking
cals is that, due to the way the device driver API/Firmware is designed, we
need to move all 4 ports to calibrated state to change a cal (even if we are
only changing cals in one specific port). Hence, we need to make sure that if
one of the ORxs is enabled, we block all the tracking calibrations controls
in all enabled ports.

Also to note that for things to work properly the glade file had to be
changed so that the ORx powerdown toggle is enabled by default.

### DIFF
--- a/plugins/adrv9002.c
+++ b/plugins/adrv9002.c
@@ -308,6 +308,85 @@ static void save_digital_gain_ctl(GtkWidget *widget, struct adrv9002_rx *rx)
 	g_free(digital_gain);
 }
 
+static void orx_control_track_cals(const struct adrv9002_common *chan, bool en)
+{
+	int i = 0;
+
+	if (!chan->enabled)
+		return;
+
+	for (i = 0; i < chan->num_widgets; i++) {
+		if (!strstr(chan->w[i].attr_name, "_tracking_en"))
+			continue;
+		gtk_widget_set_sensitive(chan->w[i].widget, en);
+	}
+}
+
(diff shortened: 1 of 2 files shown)


plugins/adrv9002.c
```diff
@@ -307,6 +307,85 @@ static void save_digital_gain_ctl(GtkWidget *widget, struct adrv9002_rx *rx)
...
+}
+
+static void orx_control_rx_widgets_visibility(const struct adrv9002_common *rx, bool en)
+{
+	char rx_str[32];
+	GtkWidget *rx_frame;
+
+	/* Just disable all the RX frame as it does not make sense to control if @en=true*/
+	sprintf(rx_str, "frame_rx%d", rx->idx + 1);
+	rx_frame = GTK_WIDGET(gtk_builder_get_object(rx->priv->builder, rx_str));
```
dNechita, line 384: It would be more bulletproof to check if rx_frame is not NULL before going forward.

dNechita (approved):
Looks good

plugins/adrv9002.c
nunojsa: I can do that :+1:

nunojsa:
V2:

* Check that `rx_frame` is not null.

PR #341, opened 2021-06-29
Misc fixes and improvements

nunojsa:
This PR is result of the issues spotted in #340 . The first patch makes the string compare function (used for sorting devices) aware of IIO labels. The rest of the series is  handling the dynamic plugin name creation issue. This is done by introducing a new helper in osc to handle `get_data_for_possible_plugin_instances()`. As a bonus, we remove duplicated code used in all plugins supporting dynamic creation.

Also to note that adrv9002 and ad9081 were not affected by the plugin name issue because they were re-sorting the list of devices returned by `get_iio_devices_starting_with()`. With the new helper, there's no need for that anymore.


dNechita (approved):
really nice!

PR #344, opened 2021-07-14
Fix osc net connection

nunojsa:
The main purpose of this is to change to the way we check for a broken connection with IIOD as that changed in libiio. In the process some misc fixes and mainly coding style improvements were done.

Adding @pcercuei as a reviewer so that he can ack if the new method is in accordance with his suggestion...

### DIFF
--- a/osc.c
+++ b/osc.c
@@ -1637,13 +1637,17 @@ static gboolean idle_timeout_check(gpointer ptr)
 {
 	int ret;
 	struct iio_context *new_ctx = (struct iio_context *) ptr;
+	struct iio_device *dev;
+	const struct iio_device *trig;
 
 	if (new_ctx != ctx)
 		return FALSE;
 
-	/* We use iio_context_get_version here just as a way to ping the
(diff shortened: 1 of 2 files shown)


osc.c
```diff
@@ -1637,13 +1637,16 @@ static gboolean idle_timeout_check(gpointer ptr)
 {
 	int ret;
 	struct iio_context *new_ctx = (struct iio_context *) ptr;
+	/* Just get the first dev. We only run this if there are devices in the IIO context */
+	struct iio_device *dev = iio_context_get_device(new_ctx, 0);
```
dNechita, line 1641: Wouldn't it make sense to get device after this check: if (new_ctx != ctx)? Just in case ptr is NULL or even if new_ctx is just different from ctx.

nunojsa: I think in the end we'll be fine like this :). But I will change to your suggestion

nunojsa:
v2:

* Get iio device after checking valid IIO context.

dNechita (approved):

PR #345, opened 2021-07-14
add iio label support for adrv9002 

nunojsa:
This adds support for IIO labels where needed for the adrv9002 eval boards. I first though I would grep for `iio_device_get_name()` and replace with the new helper introduced here but then I saw that maybe we do not need it in all places. Hence, let's introduce IIO labels "on demand".

Note that this will probably have to syn with #344

### DIFF
--- a/dialogs.c
+++ b/dialogs.c
@@ -24,6 +24,7 @@
 #include "osc.h"
 #include "config.h"
 #include "phone_home.h"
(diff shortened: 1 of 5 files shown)


dNechita (approved):

nunojsa:
v2:

* rebased on top of master

PR #346, opened 2021-07-15
Fix CentOS 8 Azure builds. CMake must grab from kitware and doxygen m…

ccraluca:
For CentOS 8:
-CMake has a bug when installed from repos. Must grab from kitware
-doxygen must be installed with "dnf", not "yum"

Signed-off-by: Raluca Chis <raluca.chis@analog.com>


nunojsa (commented):
I would only argue against the commit message. It is unwrapped... I would prefer to see a clear subject like "ci: fix centOs Azure builds" and then give a proper description of what is being fixed in the message

ccraluca:
Ok, I will reformulate

nunojsa (approved):

dNechita (approved):

PR #349, opened 2021-08-06
Support cmake mingw based builds

nunojsa:
The main goal of this PR is to support building osc with cmake using mingw. Hence, the CI scripts for creating the Windows package can be updated to use cmake. Along the way, some refactoring was also made. The ultimate goal (when this is merged) is to __finally__ get rid of the legacy Makefile.


nunojsa:
Reviews are already welcomed. I just converted this to draft since I still want to test at least a running plugin both on Linux and Windows. I did tested that osc at least starts correctly on both platforms...

nunojsa:
Ok, I did a quick test on the adrv9002 eval board which uses the adrv9002, axi-tdd, dmm and debug plugins... Things seem to work so I will remove this PR from draft

dNechita (approved):
This looks good.

PR #352, opened 2021-08-11
plugins/spectrum_analyzer: Add cmake option to build it or not

dNechita:
And by default exclude it from build.


mhennerich (commented):
Wondering, if we're not building and installing the plugin anymore how can the ifdef in identify() do anything?
The original problem I see is that there is no clean uninstall when a new OSC is build and installed.
So all residual plugin SOs will get loaded. the only way to prevent this is to clean the plugin folder. Or to update the plugin with this identify hack.

dNechita:
> Wondering, if we're not building and installing the plugin anymore how can the ifdef in identify() do anything?
> The original problem I see is that there is no clean uninstall when a new OSC is build and installed.
> So all residual plugin SOs will get loaded. the only way to prevent this is to clean the plugin folder. Or to update the plugin with this identify hack.

Makes sense. Initially I kept building the plugin but then I modified. I will make the changes.

CMakeLists.txt
```diff
@@ -302,4 +305,9 @@ foreach(plib_dest xmls filters waveforms block_diagrams)
 	install(DIRECTORY ${plib_dest} DESTINATION ${PLIB_DEST})
 endforeach()
 
+if (WITH_PLUGIN_SPECTRUM_ANALYZER)
+	add_definitions(-DPLUGIN_SPECTRUM_ANALYZER)
+endif()
+
+
```
nunojsa, line 312: I guess this option should be in the plugins specific CMakeLists file. Moreover, I would just not compile the library at all if the option is OFF. As it is the default value, I would remove the plugin from the list of plugins and just add it `if (WITH_PLUGIN_SPECTRUM_ANALYZER)`

nunojsa:
> Wondering, if we're not building and installing the plugin anymore how can the ifdef in identify() do anything?
The original problem I see is that there is no clean uninstall when a new OSC is build and installed.
So all residual plugin SOs will get loaded. the only way to prevent this is to clean the plugin folder. Or to update the plugin with this identify hack.

IMO, I would just vote for the first option "clean the plugin folder"...  Maybe adding a [uninstall target](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake)?

CMakeLists.txt
dNechita: I did that initially but as Michael said we need to build the plugin. As for the option, I prefer all options to be at top level cmake.

nunojsa: Cannot say I agree with that but fair enough...

tfcollins (approved):

PR #354, opened 2021-09-22
[2019_R2] Adrv9002 new sdk

nunojsa:
Update the adrv9002 plugin to the new SDK integrated in the linux driver


plugins/adrv9002.c
```diff
@@ -281,7 +281,7 @@ static void save_digital_gain_ctl(GtkWidget *widget, struct adrv9002_rx *rx)
 	combo_box_save(widget, &rx->digital_gain_ctl);
 	digital_gain = gtk_combo_box_text_get_active_text(GTK_COMBO_BOX_TEXT(widget));
 
-	if (digital_gain && strstr(digital_gain, "automatic_control")) {
+	if (digital_gain && strcmp(digital_gain, "spi")) {
```
dNechita, line 284: Do we want to keep a backward compatibility? (look for automatic_control in case spi isn't found?)

dNechita (approved):

plugins/adrv9002.c
nunojsa: Not really... The old value was completely removed from the driver even in the release branch. It was removed because it was wrong so I did not bothered in keeping compatibility with a buggy ABI :)

PR #371, opened 2022-05-04
plugins:dac_data_manager: Add support for devices without DDS

rbolboac:
In case the DAC device does not offer DDS support, but it offers buffer capabilities, show DAC buffer settings.

### DIFF
--- a/plugins/dac_data_manager.c
+++ b/plugins/dac_data_manager.c
@@ -1802,6 +1802,9 @@ static void manager_iio_setup(struct dac_data_manager *manager)
 	GSList *node;
 	guint i;
 
+	if (!manager->tones_count)
+		return;
+
 	for (node = manager->dds_tones; node; node = g_slist_next(node))
 		tone_setup(node->data);
 
@@ -1894,8 +1897,6 @@ static void dds_non_iq_tx_init(struct dds_dac *ddac, struct dds_tx *tx, unsigned
 static int dds_dac_init(struct dac_data_manager *manager,
 	struct dds_dac *ddac, struct iio_device *iio_dac)
 {
-	int ret = 0;
-
 	if (!iio_dac)
 		return 0;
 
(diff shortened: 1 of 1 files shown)


dNechita (approved):
Looks good!

plugins/dac_data_manager.c
```diff
@@ -2035,13 +2036,15 @@ struct dac_data_manager *dac_data_manager_new(struct iio_device *dac,
 void dac_data_manager_free(struct dac_data_manager *manager)
 {
 	if (manager) {
+		if (manager->tones_count) {
 		free(manager->dac1.txs);
 		free(manager->dac2.txs);
+		g_slist_free(manager->dds_tones);
```
nunojsa, line 2042: the indentation here is wrong...

plugins/dac_data_manager.c
```diff
@@ -1927,7 +1927,7 @@ static int dds_dac_init(struct dac_data_manager *manager,
 	manager->dacs_count++;
 	ddac->index = manager->dacs_count;
 
-	return ret;
+	return 0;
```
nunojsa, line 1930: As this api stands now, there's no point in returning anything... I'm not seeing a path where an error is set. Hence, I'm also not seeing the point of the `fprintf()`...

In fact, I'm wondering if we should have a better way of detecting such devices... Like this, we loose the ability to say that something is wrong for devices that are capable of more than 1 tone

rbolboac:
- fixed indentation 
- changed dds_dac_init

plugins/dac_data_manager.c
dNechita: yes, you are right that we need a better way of detecting devices that do not have DDS.
One criteria would be that the number of tones is equal to 0.

rbolboac: Thank you for pointing this out!
I made some changes. Do you think the new variant is more suitable?

nunojsa (approved):

PR #380, opened 2022-07-04
plugins: cn0511: remove dac frequency set

amiclaus:
Do not force setting the dac frequency to 6GHz.

The frequency is set now to 6GHz at power-up with:
https://github.com/analogdevicesinc/linux/pull/1941

Signed-off-by: Antoniu Miclaus <antoniu.miclaus@analog.com>


amiclaus:
Since the overlay is currently loaded via EEPROM, closing this PR to make sure that all boards that have the old overlay will work properly.

PR #383, opened 2022-07-12
CMakeLists: constrain plugins install

amiclaus:
The plugins that are dependent on libad9361 and libad9166 are skipped if
the libraries are not found, sending a warning message to the user.

Signed-off-by: Antoniu Miclaus <antoniu.miclaus@analog.com>


CMakeLists.txt
```diff
@@ -138,6 +138,16 @@ else()
...
 
+if ((NOT LIBAD9166_LIBRARIES) AND (NOT LIBAD9166_INCLUDE_DIRS))
+	set(LIBAD9166_LIBRARIES "")
+	set(LIBAD9166_INCLUDE_DIRS "")
+endif()
+
+if ((NOT LIBAD9361_LIBRARIES) AND (NOT LIBAD9361_INCLUDE_DIRS))
+	set(LIBAD9361_LIBRARIES "")
+	set(LIBAD9361_INCLUDE_DIRS "")
+endif()
```
nunojsa, line 149: Do we need to do this extra step? Can't we do `if (NOT LIBAD9361_LIBRARIES)` and print the message plus the removal?

plugins/CMakeLists.txt
```diff
@@ -55,6 +55,17 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
...
 endif()
 
+if("${LIBAD9166_LIBRARIES}" STREQUAL "")
+	message(WARNING "libad9166 not found. The following plugins will not be installed: cn0511")
+	list(REMOVE_ITEM PLUGINS cn0511)
+endif()
+
+if("${LIBAD9361_LIBRARIES}" STREQUAL "")
+	message(WARNING "libad9361 not found. The following plugins will not be installed: fmcomms2
+		fmcomms2_adv spectrum_analyzer")
```
nunojsa, line 65: nit: maybe put the plugins dependent on this libraries in a variable so that it's easier to handle if we ever need to extend the list or use that list anywhere else

nunojsa (commented):
minor comments from my side. Other than that, LGTM

plugins/CMakeLists.txt
amiclaus: I agree.

CMakeLists.txt
amiclaus: If we don't do:
```
set(LIBAD9166_LIBRARIES "")
set(LIBAD9166_INCLUDE_DIRS "")
```
the error will persist:
`CMake Error: The following variables are used in this project, but they are set to NOTFOUND.`

with the current structure of the CMakeLists.txt

I wanted to keep the same approach as for libserialport:
https://github.com/analogdevicesinc/iio-oscilloscope/blob/73a5a40a4c720c88650ef0dad7a77214a5d493e8/CMakeLists.txt#L130-L139

nunojsa: Ahh I see this is needed because of `set(LIBOSC_LIBS...)`

amiclaus (commented):
changes in v2:
 - add variables for libad9166 and libad9361 dependent plugins.

mhennerich (approved):

plugins/CMakeLists.txt
```diff
@@ -50,11 +50,29 @@ set(PLUGINS
...
+)
+
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
 	list(REMOVE_ITEM PLUGINS spectrum_analyzer scpi)
 	set(EXTRA_WIN_LIBRARIES winpthread)
 endif()
 
+if("${LIBAD9166_LIBRARIES}" STREQUAL "")
+	message(WARNING "libad9166 not found. The following plugins will not be installed: ${LIBAD9166_PLUGINS}")
+	list(REMOVE_ITEM PLUGINS cn0511)
```
nunojsa, line 68: `${LIBAD9166_PLUGINS}`?

amiclaus: true, thanks for the catch :)

amiclaus (commented):
v3:
- use LIBAD9166_PLUGINS variable when removing items from list instead of specific plugin name.

nunojsa (approved):

PR #393, opened 2022-12-13
plugins/dac_data_manager: First, assume buffer is cyclic

dNechita:
This should be default.


rbolboac (approved):
Thank you for fixing this!
Lgtm

nunojsa (approved):
tested with adrv9002 plugin...

mhennerich (approved):

PR #394, opened 2022-12-13
cmake: Move -rdynamic flag from compiler to linker

dNechita:
This is because rdynamic is a linker flag. And also there are reports that during build warnings(or errors) happend because of this flag being provided to compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,7 +201,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
 	add_compile_options(-mwindows)
 	set(WIN_LINK_FLAGS "-Wl,--export-all-symbols -Wl,--subsystem,windows")
 else()
-	add_compile_options(-rdynamic)
+	add_link_options(-rdynamic)
 endif()
 
 add_library(osc ${OSC_SRC})


CMakeLists.txt
```diff
@@ -201,7 +201,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
 	add_compile_options(-mwindows)
 	set(WIN_LINK_FLAGS "-Wl,--export-all-symbols -Wl,--subsystem,windows")
 else()
-	add_compile_options(-rdynamic)
```
nunojsa, line 204: I would say we can still use `add_compile_options()`... What `-Wl,option` means is that option is to be passed to the linker...

dNechita: ok

dNechita:
v2:

- continue using add_compile_options()

CMakeLists.txt
nunojsa: Hi Dan,

Well, sorry but my comment is not correct as the option is still passed during the compilation phase (even though only used by the linker). So, IIRC. your original patch was like this, right?

```
[nsa@nsa build]$ git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9118446efa46..b50280464baa 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,7 +201,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
        add_compile_options(-mwindows)
        set(WIN_LINK_FLAGS "-Wl,--export-all-symbols -Wl,--subsystem,windows")
 else()
-       add_compile_options(-rdynamic)
+       add_link_options(-rdynamic)
 endif()

 add_library(osc ${OSC_SRC})

```
Note that we still have issues when compiling with clang... But those can be fixed in a different PR

dNechita:
V3:

- switch back to initial changes which are using add_link_options()

CMakeLists.txt
dNechita: Hi Nuno,
No worries. Thanks for testing this and getting back to me.

nunojsa (approved):

PR #395, opened 2022-12-15
plugins: adrv9002: fix segfault with dac buffering

nunojsa:
The DAC Manager holds a weak reference to the IIO context (passed by the plugin) which it uses to handle the dac devices (eg: creating buffers). In dac_data_manager_free(), the Manager will destroy any buffers that were created which means we cannot destroy the IIO context before calling this function...

While at it, reordered how resources are freed during the plugin destructor so that we free things in the reverse order they were created (as it should be typically done).


nunojsa:
v2:

* Also fix the ad9081 plugin.

dNechita (approved):
Looks good. Thanks for investigating and fixing it.

PR #409, opened 2023-02-09
clang fixes

nunojsa:
Needed to compile with clang. Both of the "fixes" make sense. 

It might make sense to have a CI job to also compile things with clang. Compiling with more than one compiler is always a good thing as it allow us to catch more things.


dNechita (approved):
looks good

dNechita:
I also added an issue (#411) to don't forget about your suggestion. Thanks!

pcercuei (approved):

PR #413, opened 2023-02-14
Gtk3 Migration

cristina-suteu:
The application will now use GTK3 as base. Major changes include:
- use GtkGrid instead of GtkTable
- remove deprecated widgets
- use stylesheet and CSS provider for plot background
- set align properties instead of using Gtk Alignment widget
- remove create_frame_with_table method and create widgets manually
- replace _MINGW32_ flag for windows compilation with generic _MINGW_ flag

In order to produce windows installers, the following changes were implemented: 
- created windows docker image with : msys64 + all osc dependencies installed
- added github actions flow which pulls and runs the docker image in order to build the application and produce windows installers on Push and on Pull Requests
- remove unused appveyor files

This method of building the application is more straightforward and less error prone than cross-compiling.


mhennerich (approved):

nunojsa (approved):
There's just one thing that is still not ideal which are the glade files for the various plugins... They are still for gtk2 (at least adr9002). Just opening them with a newer glade and saving them provides some changes that will set these files "compatible" with gtk3 (at least glade does not complain anymore). However, when saving the files, glade complains about some errors.

Anyways, I guess we can move/fix things on a plugin by plugin basis after merging this...

dNechita (approved):

PR #414, opened 2023-02-15
CMakeLists: min. version increased to 3.13

cristina-suteu:
The CMake version needs to be bumped to 3.13 in order to support the add_link_options CMake command. 
This will also eliminate the CMake Deprecation Warning.


pcercuei:
Why do we want to increase the min CMake version to 3.13? Your commit doesn't say :)

nunojsa (approved):

nunojsa:
Yeah, if we want to be extra paranoid maybe it makes sense... FWIW, If it was only up to me, I would probably just merge it already :)

Issue #404, opened 2023-02-07
Create .deb packages


dNechita:
It would be nice if we could generate .deb packages for IIO-Oscilloscope.
It could be done once for each release (manually or automated). Additionally, the package could be generated for each CI build job.

gastmaier:
I created a [pull request to generate flatpaks](https://github.com/analogdevicesinc/iio-oscilloscope/pull/421).
About the debian package, iio-oscilloscope is dynamically linked to these libraries:

<details>
<summary>
shared library dependencies 
</summary>

```bash
[📦 org.adi.IIO-Oscilloscope lib]$ ldd libosc.so
	linux-vdso.so.1 (0x00007fff9596b000)
	libglib-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f5744a7a000)
	libgtk-3.so.0 => /usr/lib/x86_64-linux-gnu/libgtk-3.so.0 (0x00007f5744200000)
	libgdk-3.so.0 => /usr/lib/x86_64-linux-gnu/libgdk-3.so.0 (0x00007f57440f2000)
	libgdk_pixbuf-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgdk_pixbuf-2.0.so.0 (0x00007f5744a4b000)
	libgobject-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007f57449e8000)
	libgtkdatabox.so.1 => /app/lib/libgtkdatabox.so.1 (0x00007f57440cc000)
	libfftw3.so.3 => /usr/lib/x86_64-linux-gnu/libfftw3.so.3 (0x00007f5743ed0000)
	libiio.so.0 => /app/lib/libiio.so.0 (0x00007f5743ead000)
	libxml2.so.2 => /app/lib/libxml2.so.2 (0x00007f5743d3d000)
	libcurl.so.4 => /usr/lib/x86_64-linux-gnu/libcurl.so.4 (0x00007f5743c9a000)
	libjansson.so.4 => /app/lib/libjansson.so.4 (0x00007f5743c89000)
	libmatio.so.11 => /app/lib/libmatio.so.11 (0x00007f5743c18000)
	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x00007f5743a00000)
[...]

PR #422, opened 2023-03-21
 Create .deb package target

gastmaier:
Generate a deb package, as we already do with libiio.

It is a copy from how libiio generates the deb package, with updated dependencies and a special case for GtkDataBox [1].
It should support Ubuntu version >= 20.04 .

I did not really like [libiio's way of generating the debs/rpms](https://github.com/analogdevicesinc/libiio/blob/master/cmake/LinuxPackaging.cmake), but for cohesion I used the same script.
That being said, I don't think we should ever enable DEB_DETECT_DEPENDENCIES to auto-detect dependencies versions.

The CI is missing, but will probably be very similar to the one in libiio.

[1] In #404 it was suggested to generate debs for the Gtk2 version, but that means to branch from a old commit and create a "legacy" release; instead, I bundled it (140kB) in the deb/rpm (all of its deps are covered by IIO-Oscilloscope dependencies).

Suggestions are welcome.


gastmaier:
I included the CI with commit 2ed30f7, however libad*-iio is not included in the main  Ubuntus repo.
I was thinking to bundle with iio-osc, instead of making the user install 2 more packages.

gastmaier:
Summary of the rebase (5 days ago):
- Supports Ubuntu 22.04 (v0.15); for Ubuntu 20.04 (v0.14), there is branch [ubuntu20.04-v0.14](https://github.com/analogdevicesinc/iio-oscilloscope/tree/ubuntu20.04-v0.14).
- Uses only packages from the main Ubuntu repository as dependencies.
- Checks if libad9361 or libad9166 is included, but since only libad9361 is available in the main Ubuntu repository, it is not compiled with libad9166 support.

dNechita (commented):
Looks good. I will give it a try today or tomorrow and get back to this PR. Thanks!

gastmaier:
Requires changes to meet new apt repo requirements + wait pkg deps to be stable in the apt.

PR #424, opened 2023-04-24
UI fixes

cristina-suteu:
This PR fixes some of the known UI issues :
- updates glade files for gtk3  
- ensures all plugin tabs expand in the main window
- fixes the math channel menu in osc-plot
- in dac_data_manager, tones are aligned by column
- fmcomms2 plugin: fix spin buttons to show the entire number


dNechita:
what version of Glade did you use to open, save & close the .glade files?

glade/spectrum_analyzer.glade
```diff
@@ -94,6 +95,7 @@
                     <property name="visible">True</property>
                     <property name="can-focus">True</property>
                     <property name="invisible-char">•</property>
+                    <property name="width-chars">10</property>
```
nunojsa, line 98: just a nit about this one. You eventually mention that other plugins should care about this but your commit message is misleading in the sense that it looks like you're doing this for all plugins which is not the case. Hence, I would do one of two:

* Either update all plugins in one patch;
* Divide the current patch in 2 for each individual plugin you're addressing.

nunojsa (approved):

cristina-suteu:
@dNechita I used glade 3.38, installed via apt-get on ubuntu 22

glade/spectrum_analyzer.glade
cristina-suteu: I divided the patch in 2, thanks

PR #426, opened 2023-04-25
ad9739a.c & dac_data_manager.c: Change name from "hpc" to "lpc"

IuliaCMoldovan:
Linked to [ADI Linux PR 2125](https://github.com/analogdevicesinc/linux/pull/2125).

### DIFF
--- a/plugins/dac_data_manager.c
+++ b/plugins/dac_data_manager.c
@@ -841,7 +841,7 @@ static GtkWidget *gui_dds_mode_chooser_create(struct dds_tx *tx)
 	GtkWidget *box;
 	GtkWidget *dds_mode_lbl;
 	GtkComboBoxText *dds_mode;
-	bool no_buffer_support = !strcmp(tx->parent->name, "axi-ad9739a-hpc");
+	bool no_buffer_support = !strcmp(tx->parent->name, "axi-ad9739a-lpc");
 
 	box = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 10);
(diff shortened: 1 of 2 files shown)


dNechita:
Would it make sense to make this backwards compatible? Meaning that for whoever uses some old designs and still has -lpc suffix this plugin will still work. Or is the design so old that no one will use?

plugins/dac_data_manager.c
```diff
@@ -841,7 +841,11 @@ static GtkWidget *gui_dds_mode_chooser_create(struct dds_tx *tx)
 	GtkWidget *box;
 	GtkWidget *dds_mode_lbl;
 	GtkComboBoxText *dds_mode;
-	bool no_buffer_support = !strcmp(tx->parent->name, "axi-ad9739a-hpc");
+	bool no_buffer_support=false;
+	if(!strcmp(tx->parent->name,"axi-ad9739a-lpc"))
+	  no_buffer_support = !strcmp(tx->parent->name, "axi-ad9739a-lpc");
```
dNechita, line 846: Well, you can directly set no_buffer_support to true. You already did the comparison in the if statement.
You can even go further and do something like:
if (name is axi-ad9739a-lpc or is axi-ad9739a-hpc) then set no_buffer_suport to true. Which I believe is more readable.

cristina-suteu: Fixed

cristina-suteu:
This PR is pending hardware testing

dNechita (approved):

PR #436, opened 2023-06-22
Update Developer Note

cristina-suteu:
With the migration to GTK3 completed, using Glade for GTK2 is no longer necessary. 
Therefore, when creating new glade files with Glade versions newer than 3.8, running the fix-up scripts is not required. From now on, we recommend using Glade for GTK3 when creating new glade files (recommended version v3.38).


cristina-suteu:
@nunojsa dev-tools is indeed no longer needed, so I deleted the folder.

dNechita (approved):

PR #449, opened 2023-08-22
CI:add new builds

ccraluca:
Add bullseye and rpi4 builds on Azure CI.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -117,6 +117,10 @@ stages:
           image: tfcollins/libiio_ubuntu_18_04-ci-arm-ppc:latest
           arch: aarch64
           artifactName: 'Ubuntu-arm64v8'
+        bullseye:
+          image: tfcollins/libiio_debian_bullseye-ci:latest
+          arch: x86_64
+          artifactName: 'Linux-Debian-11'
     steps:
     - checkout: self
       fetchDepth: 1
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -94,6 +94,63 @@ stages:
 
   #######################################
 
+  - job: RPIBuilds
+    pool:
+      name: Default
+      demands:
+        - agent.name -equals iio-osc_rpi
+    workspace:
+      clean: all
+    steps:
+    - checkout: self
+      fetchDepth: 1
+      clean: true
+      persistCredentials: true
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        source: 'specific'
+        project: '$(System.TeamProjectId)'
+        pipeline: $(libiioPipelineId)
+        artifact: 'Ubuntu-arm32v7'
+        runVersion: 'latestFromBranch'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -117,6 +117,10 @@ stages:
           image: tfcollins/libiio_ubuntu_18_04-ci-arm-ppc:latest
           arch: aarch64
           artifactName: 'Ubuntu-arm64v8'
+        bullseye:
```
SRaus, line 177: Can you change name to have the same format: "debian_bullseye_arm64v8"?
And I think you can add a build for 32bit image  just by specifying "arch: arm"

azure-pipelines.yml
```diff
@@ -94,6 +94,63 @@ stages:
...
+    - checkout: self
+      fetchDepth: 1
+      clean: true
+      persistCredentials: true
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        source: 'specific'
+        project: '$(System.TeamProjectId)'
+        pipeline: $(libiioPipelineId)
+        artifact: 'Ubuntu-arm32v7'
```
SRaus, line 114: Why are you using artifacts from Ubuntu inside RPI build (line #114, #124, #134 etc)?
Even if that would work, it is not ok.

azure-pipelines.yml
```diff
@@ -94,6 +94,63 @@ stages:
 
   #######################################
 
+  - job: RPIBuilds
+    pool:
+      name: Default
```
SRaus, line 99: Is it ok to let the agent connected to "Default" pool?

I'm mean, if there are other jobs that are using "Default Pool" for building things, but without specifying an explicit agent, the build may go on this.

azure-pipelines.yml
```diff
@@ -94,6 +94,63 @@ stages:
...
+        runBranch: 'refs/heads/master'
+        path: '$(Agent.BuildDirectory)/s/download/'
+      displayName: 'Get libad9166-iio artifacts'
+    - script: |
+        sudo dpkg -i ./download/*.deb
+      displayName: 'Install Dependencies'
+    - script: |
+        mkdir build && cd build
+        cmake ../
+        make -j9
```
SRaus, line 145: RPI do not have 9 CPU cores. This kind of builds will just kill it (freeze it) for at least few minutes. 
Please check how many cores its CPU has and set make command to use maximum <total number of cores minus one> threads.

azure-pipelines.yml
```diff
@@ -94,6 +94,63 @@ stages:
...
+      displayName: 'Build'
+    - task: PublishPipelineArtifact@1
+      inputs:
+        targetPath: '$(Build.ArtifactStagingDirectory)'
+        artifactName: 'RPI4-Bullseye'
```
SRaus, line 150: Please add also architecture in the name of artifacts(armhf).

SRaus (changes requested):

PR #453, opened 2023-10-05
remove error message for devices without triggers

cristina-suteu:
do not print :
" Failed to check if device has trigger. Error:No such file or directory"

The message is misleading, as there is no error. It's just that the device doesn't have a trigger

### DIFF
--- a/osc.c
+++ b/osc.c
@@ -1430,10 +1430,6 @@ static double read_sampling_frequency(const struct iio_device *dev)
 					sizeof(buf));
 			else
 				ret = -ENOENT;
-		} else {
-			fprintf(stderr, "Failed to check if device: %s has trigger. Error:"
-				"%s\n", iio_device_get_name(dev) ?: iio_device_get_id(dev),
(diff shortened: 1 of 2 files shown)


dNechita:
Please don't merge yet. I need to check something. There is a long story behind this.

dNechita:
Short story: this is OK to merge.

Long story: iio_device_get_trigger() should return -ENOENT (according to this: https://github.com/analogdevicesinc/libiio/blob/v0.25/tests/iio_info.c#L297) when the iio device doesn't need a trigger. So iio-osc could check for this error and only print the error message for other errors.
BUT there are systems that return -EIO instead of -ENOENT. And I don't know why. And I've seen at least 2-3 systems where this happens. Therefore iio-osc can't now for sure when a iio device doesn't need a trigger due to the lack of consistency mentioned above. So yes, we better hide the error. The UI interface where users can interact with trigger settings will continue to be disabled if an error is returned.

dNechita (approved):

PR #467, opened 2023-11-30
Enable window correction in FFT plot

cristina-suteu:
Added check button to enable the window correction 
check button is disabled by default
the check button state is passed as a member of the _fft_settings / _freq_spectrum_settings  structures

### DIFF
--- a/datatypes.h
+++ b/datatypes.h
@@ -143,6 +143,7 @@ struct _fft_settings {
 	struct marker_type **markers_copy;
 	GMutex *marker_lock;
 	enum marker_types *marker_type;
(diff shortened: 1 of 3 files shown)


mhennerich (commented):
Looks good! Thanks 
Two minor things:

1. The window correction should be by default enabled
2. I think we should grey out the button while in run mode, similar to the size and window type, since changing the state only takes into effect when you start/run.

cristina-suteu:
@mhennerich fixed in new commit

mhennerich (approved):

PR #474, opened 2024-01-15
osc: Implement a way to provide custom algorithm for XY plot

dNechita:
Usage example:
/* define the routine with the algorithm */
static void my_xy_transform(gfloat *x_in, gfloat *y_in, gfloat *x_out, gfloat *y_out,
	unsigned int length, void *user_data)
{
	unsigned int i;
	for (i = 0; i < length; ++i) {
		x_out[i] = some_custom_algorithm(x_in[i]);
		y_out[i] = some_custom_algorithm(y_in[i]);
	}
}

/* Look for an already created plot instance that is configured to display XY data. Then, configure it to call your routine. Make sure that the plot is stopped the first time you set the routine. This is to tell the plot to create some internal buffers first. */
bool plot_is_running = false;
OscPlot *xy_plot = plugin_find_plot_with_domain(XY_PLOT); if (xy_plot) {
	plot_is_running = osc_plot_running_state(xy_plot);
	if (!plot_is_running) {
		osc_plot_set_xy_transform(xy_plot, my_xy_transform, NULL);
	}
}

### DIFF
--- a/datatypes.h
+++ b/datatypes.h
@@ -150,6 +150,9 @@ struct _constellation_settings {
 	gfloat *x_source;
 	gfloat *y_source;
 	unsigned int num_samples;
+	void (*xy_transform)(gfloat *x_in, gfloat *y_in, gfloat *x_out, gfloat *y_out,
+			unsigned int length, void *user_data); /* optional */
+	void *xy_transform_user_data; /* optional */
 };
 
 struct _cross_correlation_settings {
--- a/oscplot.c
+++ b/oscplot.c
@@ -418,6 +418,9 @@ struct _OscPlotPrivate
 
 	void (*quit_callback)(void *user_data);
 	void *qcb_user_data;
+	void (*xy_transform)(gfloat *x_in, gfloat *y_in, gfloat *x_out,
+		gfloat *y_out, unsigned int length, void *user_data);
+	void *xy_transform_user_data;
 };
 
 G_DEFINE_TYPE_WITH_PRIVATE(OscPlot, osc_plot, GTK_TYPE_WIDGET)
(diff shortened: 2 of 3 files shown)


nunojsa (commented):
LGTM...

The only thing I see is that it would be nice to include an user of this already in this PR (as a separate commit). So it's clear we are not adding code that will never be used :)

dNechita:
> LGTM...
> 
> The only thing I see is that it would be nice to include an user of this already in this PR (as a separate commit). So it's clear we are not adding code that will never be used :)

This feature has been requested by someone on EngineerZone (https://ez.analog.com/linux-software-drivers/f/q-a/576469/does-it-possible-to-add-a-plugin-to-iio-oscilloscope-before-constellation-display) in order to modify iio-osc on their system only.
I have no idea how to showcase this feature.

nunojsa:
> This feature has been requested by someone on EngineerZone (https://ez.analog.com/linux-software-drivers/f/q-a/576469/does-it-possible-to-add-a-plugin-to-iio-oscilloscope-before-constellation-display) in order to modify iio-osc on their system only.
I have no idea how to showcase this feature.

FWIW, here it goes my 2 cents on this... IMHO, we should not be adding code to core parts of osc just because one wants to use it in some custom hack/code. It's just not how open source is supposed to work. They should do it in a way that's useful for everyone and send patches to us :). If they don't really want to do that, you already gave them an example on how it should be done, so up to them to maintain that code I would say. Anyways, it is what it is and you're the one maintaining this app so up to you in the end :)

dNechita:
> > This feature has been requested by someone on EngineerZone (https://ez.analog.com/linux-software-drivers/f/q-a/576469/does-it-possible-to-add-a-plugin-to-iio-oscilloscope-before-constellation-display) in order to modify iio-osc on their system only.
> > I have no idea how to showcase this feature.
> 
> FWIW, here it goes my 2 cents on this... IMHO, we should not be adding code to core parts of osc just because one wants to use it in some custom hack/code. It's just not how open source is supposed to work. They should do it in a way that's useful for everyone and send patches to us :). If they don't really want to do that, you already gave them an example on how it should be done, so up to them to maintain that code I would say. Anyways, it is what it is and you're the one maintaining this app so up to you in the end :)

It kind of makes sense. I got excited about this since this implementation is similar to math channels. The difference is that users can customize the algorithm through math channels using the UI and at runtime while the implementation in this PR could offer the same thing but via an API. But this has never been requested until now. And now is just for XY plot and not for all types of plot.
I will suggest the user to do a fork and continue from there.
Afterwards I will close this PR which will remain in history anyway (as a closed PR).
Thanks!

dNechita:
Closing this. See PR discussions.

PR #477, opened 2024-01-29
adrv9002 plugin fix for disabled channels

andrei47w:
The profile gen was failing to create a configuration since some attributes cannot be read if the channel is disabled

### DIFF
--- a/plugins/adrv9002.c
+++ b/plugins/adrv9002.c
@@ -2400,6 +2400,21 @@ static int profile_gen_ui_refresh(GtkButton *self, struct plugin_private *priv)
 	return 0;
 }
 
+static void profile_gen_update_channels(GtkComboBox *self, struct plugin_private *priv)
+{
+	GtkWidget *channel_frame;
+	bool channel_en;
+	unsigned long i;
+	char *ch_frames[4] = {"frame_tx1_controls", "frame_tx2_controls", "frame_rx1_controls", "frame_rx2_controls"};
+	char *ch_buttons[4] = {"cb_tx_chan1_en", "cb_tx_chan2_en", "cb_rx_chan1_en", "cb_rx_chan2_en"};
+
+	for (i = 0; i < ARRAY_SIZE(ch_frames); i++) {
+		channel_frame = GTK_WIDGET(gtk_builder_get_object(priv->builder, ch_frames[i]));
+		channel_en = gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(gtk_builder_get_object(priv->builder, ch_buttons[i])));
+		gtk_widget_set_sensitive(channel_frame, channel_en);
+	}
+}
+
 static void adrv9002_combo_box_init(struct iio_widget *combo, const char *w_str,
 				    const char *attr, const char *attr_avail,
 				    struct plugin_private *priv, struct iio_channel *chann)
@@ -3053,6 +3068,19 @@ static GtkWidget *adrv9002_init(struct osc_plugin *plugin, GtkWidget *notebook,
 		/* refresh on preset changed */
 		g_builder_connect_signal(priv->builder, "cb_preset", "changed",
 					 G_CALLBACK(profile_gen_ui_refresh), priv);
+
+		/* update channel controls sensitivity */
+		g_builder_connect_signal(priv->builder, "cb_tx_chan1_en", "toggled",
+					 G_CALLBACK(profile_gen_update_channels), priv);
+
+		g_builder_connect_signal(priv->builder, "cb_tx_chan2_en", "toggled",
+					 G_CALLBACK(profile_gen_update_channels), priv);
+
+		g_builder_connect_signal(priv->builder, "cb_rx_chan1_en", "toggled",
+					 G_CALLBACK(profile_gen_update_channels), priv);
+
+		g_builder_connect_signal(priv->builder, "cb_rx_chan2_en", "toggled",
+					 G_CALLBACK(profile_gen_update_channels), priv);
 
 	} else {
 		gtk_widget_set_sensitive(GTK_WIDGET(gtk_builder_get_object(priv->builder, "boxProfileGen")), false);
--- a/plugins/adrv9002.c
+++ b/plugins/adrv9002.c
@@ -173,120 +173,6 @@ typedef struct adrv9002_config
 	clock_config clk_cfg;
 } adrv9002_config;
 
-static adrv9002_config lte_defaults(void)
-{
-	radio_config radio_config;
-	radio_config.ssi_lanes = 2;
-	radio_config.ddr = true; // needs logic
-	radio_config.short_strobe = true;
-	radio_config.lvds = true;
-	radio_config.adc_rate_mode = 3;
-	radio_config.fdd = false;
-
-	tx_radio_channel_config tx_config[2];
-	int i;
-	for(i = 0; i < CHANNEL_COUNT; i++) {
-		tx_config[i].enabled = true;
-		tx_config[i].sample_rate_hz = 61440000;
-		tx_config[i].frequency_offset_correction_enable = false;
-		tx_config[i].analog_filter_power_mode = 2;
-		tx_config[i].channel_bandwidth_hz = 38000000;
-		tx_config[i].elb_type = 0;
-		tx_config[i].orx_enabled = false;
-
-		radio_config.tx_config[i] = tx_config[i];
-	}
-
-	rx_radio_channel_config rx_config[2];
-	for(i = 0; i < CHANNEL_COUNT; i++) {
-		rx_config[i].enabled = true;
-		rx_config[i].sample_rate_hz = 61440000;
-		rx_config[i].frequency_offset_correction_enable = false;
-		rx_config[i].analog_filter_power_mode = 2;
-		rx_config[i].channel_bandwidth_hz = 38000000;
-		rx_config[i].adc_high_performance_mode = true;
-		rx_config[i].analog_filter_biquad = false;	    // got from default cfg
-		rx_config[i].analog_filter_bandwidth_hz = 18000000; // got from default cfg
-		rx_config[i].nco_enable = false;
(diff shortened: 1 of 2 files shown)


nunojsa (changes requested):

plugins/adrv9002.c
```diff
@@ -965,7 +963,20 @@ static void adrv9002_check_channel_status(struct plugin_private *priv,
...
 	 */
 	ret = iio_channel_attr_read_double(chan->w[0].chn, "rf_bandwidth", &dummy);
-	if (ret == -ENODEV) {
+
+	if(ret < 0 && ret != -ENODEV) {
+		printf("Warning: iio channel returned an error when reading it: %d. We assume the channel is enabled\n",
+		       ret);
+	}
+
+	return ret != -ENODEV;
```
nunojsa, line 972: I would probably do `ret == 0`. I would only consider a channel enabled if we get success. For the rf_bandwidth attr to return something rather than 0 or ENODEV (don't think the driver is returning any other ret value) likely means libiio is already somehow screwed (or some issue with the libiio backend) and odds are that things are far from being reliable.

I know this is not like it was before but now that I think about it, I think it makes more sense and I guess now is the time to change it.

plugins/adrv9002.c
```diff
@@ -965,7 +963,20 @@ static void adrv9002_check_channel_status(struct plugin_private *priv,
...
+}
+
+static void adrv9002_check_channel_status(struct plugin_private *priv,
+					  struct adrv9002_common *chan,
+					  const char *gtk_str)
+{
+	if (adrv9002_channel_is_enabled(chan)) {
```
nunojsa, line 979: You should separate the commits. One to introduce the new adrv9002_check_channel_status() with the needed changes in `adrv9002_check_channel_status()` and explaining we will have more places to call that new helper in a following patch. It makes the git history better and helps in reviewing (less changes per patch)

plugins/adrv9002.c
```diff
@@ -1381,48 +1392,50 @@ static int profile_gen_config_get_from_device(struct adrv9002_config *cfg, gpoin
 		sprintf(chann_str, "voltage%d", chann);
 		struct iio_channel *tx = iio_device_find_channel(priv->adrv9002, chann_str, true);
 		if(tx == NULL) {
-			sprintf(message, "\nFailed to find channel: %s!", chann_str);
+			sprintf(message, "\nFailed to find tx channel: %s!", chann_str);
```
nunojsa, line 1395: unrelated change

plugins/adrv9002.c
```diff
@@ -1649,43 +1668,67 @@ static int profile_gen_config_set_live_device(struct adrv9002_config *cfg, gpoin
...
-				gtk_toggle_button_set_active(
-					GTK_TOGGLE_BUTTON(gtk_builder_get_object(priv->builder, widget_str)),
-					cfg->radio_cfg.rx_config[chann].frequency_offset_correction_enable);
-			}
+		// frequency_offset_correction_enable
+		sprintf(widget_str, "cb_tx_chan%d_correction", chann + 1);
+		if(reset_preset) {
+			gtk_toggle_button_set_active(
+				GTK_TOGGLE_BUTTON(gtk_builder_get_object(priv->builder, widget_str)),
+				cfg->radio_cfg.tx_config[chann].frequency_offset_correction_enable);
```
nunojsa, line 1697: Is the above related at all with `fix for disabled channels`?

plugins/adrv9002.c
```diff
@@ -1649,43 +1668,67 @@ static int profile_gen_config_set_live_device(struct adrv9002_config *cfg, gpoin
...
+
+		// frequency_offset_correction_enable
+		sprintf(widget_str, "cb_rx_chan%d_correction", chann + 1);
+		if(reset_preset) {
+			gtk_toggle_button_set_active(
+				GTK_TOGGLE_BUTTON(gtk_builder_get_object(priv->builder, widget_str)),
+				cfg->radio_cfg.rx_config[chann].frequency_offset_correction_enable);
+		}
+
+		// rf_port
```
nunojsa, line 1731: ditto...

plugins/adrv9002.c
```diff
@@ -1440,93 +1453,99 @@ static int profile_gen_config_get_from_device(struct adrv9002_config *cfg, gpoin
...
 		// tx.orx_enabled
 		ret = iio_channel_attr_read(rx, "orx_en", buf, sizeof(buf));
 		if(ret < 0) {
 			if(ret == -ENODEV) {
-				tx_config[chann].orx_enabled =
+				radio_config.tx_config[chann].orx_enabled =
 					default_cfg.radio_cfg.tx_config[chann].orx_enabled; // Temporary fix
 			} else {
-				sprintf(message, "\nFailed to get channel: %s attr: %s! error code: %d", chann_str,
+				sprintf(message, "\nFailed to get rx channel: %s attr: %s! error code: %d", chann_str,
```
nunojsa, line 1543: Also unrelated change... When are minor changes like this that is ok to put it in the same patch, you should at least mention it in the commit message.

plugins/adrv9002.c
```diff
@@ -1668,44 +1668,68 @@ static int profile_gen_config_set_live_device(struct adrv9002_config *cfg, gpoin
...
-	char *ch_types[2] = {"rx", "tx"};
-	for(ch_type = 0; ch_type < ARRAY_SIZE(ch_types); ch_type++) {
-		for(chann = 0; chann < CHANNEL_COUNT; chann++) {
-			// channel_bandwidth_hz
-			sprintf(widget_str, "cb_%s_chan%d_bw", ch_types[ch_type], chann + 1);
-			sprintf(value, "%d", cfg->radio_cfg.rx_config[chann].channel_bandwidth_hz);
-			populate_combo_box(GTK_COMBO_BOX_TEXT(gtk_builder_get_object(priv->builder, widget_str)), NULL, 0, TRUE,
-					   value);
+	// TX
+	for(chann = 0; chann < CHANNEL_COUNT; chann++) {
```
nunojsa, line 1672: Note that there's already a `ADRV9002_NUM_CHANNELS`... But ideally you would use ARRAY_SIZE(). Also for tx, there's `n_tx` to accommodate for adrv9003 which only has 1 TX.

Hence, I'm not seeing the point of CHANNEL_COUNT

plugins/adrv9002.c
```diff
@@ -2400,6 +2400,21 @@ static int profile_gen_ui_refresh(GtkButton *self, struct plugin_private *priv)
 	return 0;
 }
 
+static void profile_gen_update_channels(GtkComboBox *self, struct plugin_private *priv)
+{
+	GtkWidget *channel_frame;
+	bool channel_en;
+	unsigned long i;
+	char *ch_frames[4] = {"frame_tx1_controls", "frame_tx2_controls", "frame_rx1_controls", "frame_rx2_controls"};
+	char *ch_buttons[4] = {"cb_tx_chan1_en", "cb_tx_chan2_en", "cb_rx_chan1_en", "cb_rx_chan2_en"};
```
nunojsa, line 2409: nit: I would remove the 4...

nunojsa (dismissed):
Some nits... I'm approving the PR but I do hope that at some point you get rid of CHANNEL_COUNT.

Moreover, the plugin source is getting way too complex (and big). It might make sense to detach the profile_gen from the main .c file.

plugins/adrv9002.c
```diff
@@ -198,7 +197,7 @@ static adrv9002_config lte_defaults(void)
 	}
 
 	rx_radio_channel_config rx_config[2];
-	for(i = 0; i < CHANNEL_COUNT; i++) {
+	for(i = 0; i < 2; i++) {
```
nunojsa, line 200: ADRV9002_NUM_CHANNELS...

plugins/adrv9002.c
```diff
@@ -185,7 +184,7 @@ static adrv9002_config lte_defaults(void)
 
 	tx_radio_channel_config tx_config[2];
 	int i;
-	for(i = 0; i < CHANNEL_COUNT; i++) {
+	for(i = 0; i < 2; i++) {
```
nunojsa, line 187: ADRV9002_NUM_CHANNELS...

nunojsa (approved):

plugins/adrv9002.c
andrei47w: ADRV9002_NUM_CHANNELS is not yet defined here. The fix would be to move the profile gen structures below the other defines. Is it worth it?

nunojsa: the fix should be moving the functions below the defines. I'm even guessing that function is only called during init? If so, move it close that function and reduce it's scope...

plugins/adrv9002.c
```diff
@@ -1205,6 +1091,122 @@ static void load_profile(GtkFileChooser *chooser, gpointer data)
...
+	rx1.analog_filter_power_mode = 2;
+	rx1.channel_bandwidth_hz = 38000000;
+	rx1.adc_high_performance_mode = true;
+	rx1.analog_filter_biquad = false;	    // got from default cfg
+	rx1.analog_filter_bandwidth_hz = 18000000; // got from default cfg
+	rx1.nco_enable = false;
+	rx1.nco_frequency_hz = 0;
+	rx1.rf_port = 0;
+
+	rx_radio_channel_config rx2 = rx1;
```
nunojsa, line 1121: Couple of notes from me (just about good practices :)):

1) I would have done this move before the previous patch... Thus you would not need to still have that plain 2 in the previous patch when removing CHANNEL_COUNT.
2) You should mention in your commit message that while reducing the scope you removed the for() loop by implicit memcpy(). I was actually wondering about the loops before realizing the subtle change. If you mention it, it's then clear what's going on.

On a different note (for future refactor)... Please consider note doing things like the above function where you're returning complete objects like `adrv9002_config` (which are likely not that small). That's inefficient and, honestly, far from great C. Use pointers for things like this (likely as an argument of the function).

nunojsa (approved):

PR #485, opened 2024-03-06
make sure the sizeof argument to calloc is the second argument.

tsailer:
This PR makes sure that the sizeof argument to calloc is the second argument. New versions of gcc want to enforce this.


nunojsa (commented):
Hi Thomas,

Thanks for contributing... Can you please sign off your commit and re-push? You can just do `git commit --amend -s`. Note you need an email and user in your gitconfig.

Other than the above, LGTM

nunojsa (approved):

dNechita (approved):

Issue #480, opened 2024-02-01
OSC cannot handle multiple instances of the same chip (should use label instead of name)


MikeLooijmans:
The "name" of an iio device is the chip name, and usually isn't unique.
The DMM window correctly shows and uses labels.
The OSC plot view doesn't and gets hopelessly confused. With multiple instances of chipX, the plot window will not show this:

- chipX
- channel0
- channel1
- chipX
- channel0
- channel1

Instead the list is shown as:

- chipX
- channel0
- channel1
- channel0
- channel1
- chipX

dNechita:
Hi!
Thanks for reporting this! We will look into this.

PR #488, opened 2024-03-25
Fix plot not using device labels

dNechita:
## PR Description

- Make iio-oscilloscope plot (oscplot.c) and core (osc.c) lookup IIO DEVICES by label first. This would avoid issues where multiple IIO DEVICES with the same name exists. See https://github.com/analogdevicesinc/iio-oscilloscope/issues/480.

## PR Type

## PR Checklist
- [x] I have commented new code, particulary complex or unclear areas 
- [x] I have checked in CI output that no new warnings/errors got introduced
- [x] I have updated documentation accordingly (GitHub Pages, READMEs, etc)

### DIFF
--- a/osc.c
+++ b/osc.c
@@ -449,7 +449,7 @@ static const char * device_name_check(const char *name)
 	if (!dev)
 		return NULL;
 
-	return iio_device_get_name(dev) ?: iio_device_get_id(dev);
+	return get_iio_device_label_or_name(dev);
 }
 
 /*
@@ -1511,7 +1511,7 @@ static int capture_setup(void)
 				min_timeout = timeout;
 		}
 
-		rx_update_device_sampling_freq(iio_device_get_id(dev), freq);
+		rx_update_device_sampling_freq(get_iio_device_label_or_name(dev), freq);
 	}
 
 	if (ctx)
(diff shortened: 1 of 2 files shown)


osc.c
```diff
@@ -449,7 +449,8 @@ static const char * device_name_check(const char *name)
 	if (!dev)
 		return NULL;
 
-	return iio_device_get_name(dev) ?: iio_device_get_id(dev);
+	return iio_device_get_label(dev) ?:
+		(iio_device_get_name(dev) ?: iio_device_get_id(dev));
```
nunojsa, line 453: There's an helper already... `get_iio_device_label_or_name()`.  Using the id should be very unlikely (I think I never saw a driver not using iio_dev->name). Anyways either extend the helper or use it to make the ternary simpler :)

nunojsa (changes requested):

osc.c
dNechita: I didn't notice the helper, which is much better, yes. ID is unlikely to be used indeed and if it did, it would break a lot of stuff anyway. One example that comes in mind is when osc saves attributes for a device in .ini file. It wouldn't be reliable to save for iio:device0 which might point to a different hardware after a reboot.
Thanks for the suggestion!

dNechita:
V2:
Update everything (based on review suggestions) to use helper 'get_iio_device_label_or_name()'.

cristina-suteu (approved):

nunojsa (approved):

PR #490, opened 2024-04-19
dialogs.c: move the scanning process into its own thread.

AlexandraTrifan:
## PR Description

On MacOS there were some issues with the network iio_context scanning process. Inside libiio, some mutex locks are done during scanning. This blocks the UI thread and makes the app unusable. Moving into its own thread and creating a protected section (similar to calibration processes) should resolve this issue.

Besides these changes, iio-osc also needs some different fixes to get it working on MacOS (here are 2 dev branches on this matter by Cristina and Travis).

## PR Type

## PR Checklist

### DIFF
--- a/dialogs.c
+++ b/dialogs.c
@@ -432,7 +432,8 @@ static struct iio_context * get_context(Dialogs *data)
 	}
 }
 
-static void refresh_usb(void)
+
+static void refresh_usb_thread(void)
 {
 	struct iio_scan_context *ctxs;
 	struct iio_context_info **info;
@@ -447,6 +448,7 @@ static void refresh_usb(void)
 	bool scan = false;
 	gchar *active_uri = NULL;
 
+	gdk_threads_enter();
 	widget_set_cursor(dialogs.connect, GDK_WATCH);
 
 	if (gtk_combo_box_get_active(GTK_COMBO_BOX(dialogs.connect_usbd)) != -1) {
@@ -565,6 +567,7 @@ nope:
 		gtk_widget_set_sensitive(dialogs.connect_usbd, false);
 		/* Force a clear */
 		connect_clear(dialogs.connect_net);
+		gdk_threads_leave();
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


dialogs.c
```diff
@@ -585,6 +587,14 @@ static void refresh_usb(void)
 
 	/* Fill things in */
 	connect_clear(dialogs.connect_usb);
+	gdk_threads_leave();
```
dNechita, line 590: At line 570 there is a return. I believe gdk_threads_leave() should be called just before that return as well.

dialogs.c
```diff
@@ -585,6 +587,14 @@ static void refresh_usb(void)
...
+
+}
+
+
+static void refresh_usb(void)
```
nunojsa, line 595: I'm also a bit puzzled. The commit message mentions network backend but the code seems to "say" usb?

dialogs.c
AlexandraTrifan: I'll add this.

dialogs.c
AlexandraTrifan: The old iio-osc function responsible for scanning the context list was "refresh_usb". In fact, it handles all the IIO backends based on the available filter from GUI (network, local, usb, serial).
In order to reduce the number of changes to the iio-osc code (which I thought would bring more bugs and confusion to the table) I reused the old function name: "refresh_usb". 

If you think it's necessary, I can rename all these functions and all the calls to it to something related to context scanning.

The commit message mentions network because from my tests, scanning only blocks on network scanning. There is no issue on MacOS if we force iio-osc to only scan USB backends for example. What happens is that from the iio-osc main thread, libiio scanning gets up to these points: https://github.com/analogdevicesinc/libiio/blob/libiio-v0/dns_sd_bonjour.c#L184  https://github.com/analogdevicesinc/libiio/blob/libiio-v0/lock.c#L42 and then hangs. This seems to be due to these gdk thread guards: https://github.com/analogdevicesinc/iio-oscilloscope/blob/main/oscmain.c#L245-L261 .
Libiio context scanning works fine on MacOS outside of iio-osc (in iio_info for example).


The FMCOMMS2 advanced plugin calibration thread handles this in the same way I proposed for scanning: https://github.com/analogdevicesinc/iio-oscilloscope/blob/main/plugins/fmcomms2_adv.c#L830 .

nunojsa (approved):

dNechita (approved):

cristina-suteu (approved):

PR #492, opened 2024-05-10
adrv9002: API version mismatch fix

andrei47w:
## PR Description

- validation failed if driver was "68.8.1" and profile gen was "v68.8.1", so we now check if they contain the same version substring
- added option to load profile even with API version mismatch

## PR Type
- [x] New feature (a change that adds new functionality)

## PR Checklist

### DIFF
--- a/plugins/adrv9002.c
+++ b/plugins/adrv9002.c
@@ -2263,6 +2263,25 @@ static char *profile_gen_cli_get_api(void)
 	return version;
 }
 
+static char *strip_leading_and_trailing_nonnumeric_chars(char *string)
+{
+	int i;
+	char *str = (char *)malloc(strlen(string) * sizeof(char));
+	sprintf(str, "%s", string);
+
+	while((str[0] < '0' || str[0] > '9') && str[0] != '\0') {
+		str++;
+	}
+
+	i = strlen(str) - 1;
+	while((str[i] < '0' || str[i] > '9') && i > 0) {
+		str[i] = '\0';
+		i--;
+	}
+
+	return str;
+}
+
 static bool profile_gen_check_api(gpointer data)
 {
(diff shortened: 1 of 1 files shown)


plugins/adrv9002.c
```diff
@@ -2281,11 +2281,24 @@ static bool profile_gen_check_api(gpointer data)
 		goto err;
 	}
 
-	if(strcmp(version, supported_version) != 0) {
-		sprintf(message, "\nOnly API version %s is supported, the device uses %s!", supported_version, version);
-		profile_gen_set_debug_info(data, message);
-		free(supported_version);
-		goto err;
+	if(strstr(version, supported_version) != NULL && strstr(supported_version, version) != NULL) {
```
dNechita, line 2284: You are making this condition too permissive:
For example this will pass:
```
version = "v68.8.1"
supported_version = "8.8.1"
```
Also this will pass:
```
version = "v68.8.1"
supported_version = "8.1"
```
Also this will pass:
```
version = "v68.8.1"
supported_version = "1"
```
It would make sense to consider the minimum valid string for a version (for example: "68.8.1") and find that substring in strings like: "v68.8.1", "ver68.8.1", etc.

plugins/adrv9002.c
```diff
@@ -2263,6 +2263,25 @@ static char *profile_gen_cli_get_api(void)
 	return version;
 }
 
+static char *strip_nonnumeric_chars(char *string)
```
dNechita, line 2266: It would me more accurate to call this something like: "strip_leading_and_trailing_nonnumeric_chars" as it doesn't strip nonnumeric characters between the numeric ones (such as the dots).

dNechita (approved):

PR #495, opened 2024-06-04
fix compile errors related to use of calloc and malloc

threexc:
## PR Description

While compiling iio-oscilloscope from source on Fedora 40, I encountered numerous errors related to calloc arguments being backwards (i.e. size specified before member count), and one because malloc was used without `#include <stdlib.h>`. These two patches should fix that.

I'm not sure if/how this worked for older toolchains, so if this is an incorrect fix, please let me know and reject this.

## PR Type
- [X] Bug fix (a change that fixes an issue)

## PR Checklist
- [X] I have followed the coding standards and guidelines
- [X] I have conducted a self-review of my own code changes

### DIFF
--- a/libini2.c
+++ b/libini2.c
@@ -469,7 +469,7 @@ static struct ini_loop * ini_loop_new(char *buf_with_loop, char *loop_name)
 	}
 
 	/* Store all necessary parameters of a loop */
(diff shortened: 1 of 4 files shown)


threexc:
Oops - was on master instead of main. Closing.

PR #506, opened 2024-11-29
plugins/dac_data_manager: Create iio buffer at the right moment

dNechita:
## PR Description
The idea behind the fix is the following: due to libiio v1.x API changes the creation of the iio_buffer cannot be done too early. It needs a iio_mask and at least one channel to be enabled. The need to create the iio_buffer early was just to read one attribute about alignment, and to read the attribute only once.

The fix is to create the iio_buffer at the moment is needed and read the attribute right after the creation.

The effort for this fix started because osc was crashing at startup.

## PR Type

## PR Checklist


mhennerich (approved):
LGTM

PR #520, opened 2025-02-07
plugins/dac_data_manager.c: dynamically convert data for enabled channels.

AlexandraTrifan:
## PR Description

Instead of assuming a connected device will only have a maximum of 8 DMA channels, add a mechanism to dynamically convert data for all enabled channels.
This can be reproduced with a setup of ADRV9009 and FMCOMMS8 with all 8 TX channels enabled. Loading any file with all channels enabled will display "Waveform loaded successfully" but the DMA output will be 0. Doing the same with only 4 TX channels enabled will work properly. 

## PR Type

## PR Checklist

### DIFF
--- a/plugins/dac_data_manager.c
+++ b/plugins/dac_data_manager.c
@@ -456,12 +456,10 @@ static int analyse_wavefile(struct dac_data_manager *manager,
 
 			*count = size * tx_channels * 2;
 
-			unsigned long long *sample = *((unsigned long long **) buf);
-			unsigned int *sample_32 = *((unsigned int **) buf);
 			unsigned short *sample_16 = *((unsigned short **) buf);
+			struct _complex_ref *tx_data = calloc(tx_channels, sizeof(struct _complex_ref));
 
-			struct _complex_ref tx_data[4] = {{NULL, NULL}, {NULL, NULL}, {NULL, NULL}, {NULL, NULL}};
-			mat_complex_split_t *complex_data[4];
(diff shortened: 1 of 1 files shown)


plugins/dac_data_manager.c
```diff
@@ -479,44 +477,28 @@ static int analyse_wavefile(struct dac_data_manager *manager,
...
-						    ((unsigned long long) convert(scale, tx_data[3].re[i], offset) << 32) |
-							((unsigned long long) convert(scale, tx_data[2].im[i], offset) << 16) |
-							((unsigned long long) convert(scale, tx_data[2].re[i], offset) << 0);
-					sample[j++] = ((unsigned long long) convert(scale, tx_data[1].im[i], offset) << 48) |
-						    ((unsigned long long) convert(scale, tx_data[1].re[i], offset) << 32) |
-							((unsigned long long) convert(scale, tx_data[0].im[i], offset) << 16) |
-							((unsigned long long) convert(scale, tx_data[0].re[i], offset) << 0);
+			int ch = 0;
+			int sample_i = 0;
+			int tx_data_end = (tx_channels % 2 == 0) ? (tx_channels/2) : tx_channels;
```
dNechita, line 482: Just a small thing: you could consider variables `ch`, `sample_i `and `tx_data_end `as unsigned. These variables won't be holding any negative values as it doesn't make sense. Therefore, I would make the unsigned or size_t (which is pretty much the same thing).

dNechita (approved):
This looks good!

Issue #493, opened 2024-05-29
Linux startup problem


joezhouchenye:
Installed following the instructions at https://wiki.analog.com/resources/tools-software/linux-software/iio_oscilloscope.

Build and install with the latest commit 62fd972 in the main branch.

The application has startup problems.

First, the libraries are installed at `/usr/local/lib` and I need to add LD_LIBRARY_PATH for the root user. This app running with normal users cannot detect the PLUTO device plugged in through USB.

Second, it doesn't properly disconnect the device. If I connect the PLUTO device once and quit the app, I won't be able to reopen the app again and a seg fault is raised. I will have to unplug the PLUTO, open the app, and plug the PLUTO in again to get it to work.

cristina-suteu:
Hello, 

There are 2 separate issues to be addressed here : 

1. The Segmentation Fault issue - this is a libiio v0.25 bug. This was fixed in the next release. You can download the latest libiio release from here : https://github.com/analogdevicesinc/libiio/releases/tag/v0.26 Alternatively , you can build the libiio-v0 branch. (https://github.com/analogdevicesinc/libiio/tree/libiio-v0)

2. If you want to change the installation path , there is no need to modify the CMakeLists.txt source. When running cmake , you can pass the option : -DCMAKE_INSTALL_PREFIX="your/path/". This will overwrite the path specified in CMakeLists.txt . 

I hope this helps.

PR #527, opened 2025-03-10
#493: Linux startup problem

achengli:
CMakeLists.txt library installation directory was changed from /usr/local/lib to /lib because most of the Linux distributions don't include /usr/local/lib on LD_PATH

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,8 +251,15 @@ set_target_properties(oscmain PROPERTIES OUTPUT_NAME osc)
 
 # Fix libosc.0.so not found
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-  set(CMAKE_INSTALL_LIBDIR "/lib")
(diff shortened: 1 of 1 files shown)


achengli:
I did some changes that do not apply for these architectures. I guess the problem is the project is not compatible with failing architectures.

cristina-suteu (commented):
If you want to change the installation path , there is no need to modify the CMakeLists.txt source. When running cmake , you can pass the option : -DCMAKE_INSTALL_PREFIX="your/path/". This will overwrite the path specified in CMakeLists.txt. On linux distros, the location /usr/local/lib contains libraries for locally installed programs(i.e. programs that you built from source), ensuring they do not interfere with the system's standard library.  On the other hand, the /lib location contains essential shared libraries and kernel modules needed to boot the system and run commands in the root filesystem.

achengli:
Thank you @cristina-suteu !

PR #528, opened 2025-03-14
[2023_R2] Revert "filters: adrv9002: add profiles for sdk 27.1"

nunojsa:
This reverts commit 9b4291f45ea52c04cebc569704135b60c27439dd.

As it turns out, API 68.14.13 has no meaningful fixes and this release only affects users using hdl that we do not use in the linux driver. Hence this would only raise unnecessary questions without any added value and so it was reverted from the driver. Therefore, let's revert it here as well.

## PR Type
- [ ] Bug fix (a change that fixes an issue)

## PR Checklist


dNechita (approved):

nunojsa:
> I guess it should be reverted on the main branch as well, right?

For now, I'm keeping it on linux main... Reason for dropping in the release branch is because it might raise more questions and since it would also be part of the next kuiper patch release for 2023_R2. So we decided to drop this from 2023_R2 but for main I do not think it's that important (supposed to be a dev branch)

PR #570, opened 2026-05-04
Fix profile dialog and improve URI handling

nunojsa:
## PR Description

 - Fix the Load/Save Profile dialog appearing twice on every click by wiring the signal handler in code rather than through Glade auto-connection.
 - Refactor how profile URIs are resolved during startup and profile loading. Previously, loading any profile would unconditionally reload the application — even when already connected to the matching URI — and command-line URIs (-u) could be silently overridden by a saved profile. The new handle_profile_load_request() compares URIs before
deciding whether a reload is actually needed.

@dNechita, note that I completely removed [this code](https://github.com/analogdevicesinc/iio-oscilloscope/blob/main/osc.c#L2501) given I'm not sure why it's here but it is clear that, in it's current form, it has lots of issues:

* Getting the URI twice;
* This hidden app reloading ;
* Even after getting a valid URI, we do an IIO scan and do all of that USB related stuff

So If you think we still need all of that USB stuff let me know, so we can see how can bring it back in a sane way.

## PR Type
- [ ] Bug fix (a change that fixes an issue)

## PR Checklist

### DIFF
--- a/dialogs.c
+++ b/dialogs.c
@@ -997,7 +997,7 @@ G_MODULE_EXPORT void cb_show_about(GtkButton *button, Dialogs *data)
 	gtk_widget_hide(data->about);
 }
 
-G_MODULE_EXPORT void load_save_profile_cb(GtkButton *button, Dialogs *data)
+void load_save_profile_cb(GtkButton *button, Dialogs *data)
 {
 	/* Save as Dialog */
 	gint ret;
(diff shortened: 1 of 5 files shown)


dNechita:
I think this is ok to be merged.

I did take a look at the code you removed and it seems it tries find if the PID,VID and Serial Number written in the profile file matches a scanned context that has the same PID, VID, Serial Number regardless of the bus name (e.g. usb:3.25).
But testing things showed me that the URI written in the profile file is just the usb bus name. So, the code you removed isn't used.

dNechita (approved):

Issue #10, opened 2015-04-15
osc crashes when running test.ini files


rgetz:
When I add something to an ini file, which looks like:

[IIO Oscilloscope - Capture Window3]
cycle = 200
save_markers = log.txt
quit = 1

it crashes:
Program received signal SIGSEGV, Segmentation fault.
0xb6bebf9c in iio_context_get_device () from /usr/lib/libiio.so.0
(gdb) bt
#0  0xb6bebf9c in iio_context_get_device () from /usr/lib/libiio.so.0
#1  0xb6fc56cc in device_list_treeview_init (plot=0x5e09a8) at oscplot.c:3245
#2  create_plot (plot=0x5e09a8) at oscplot.c:6172
#3  osc_plot_init (plot=0x5e09a8) at oscplot.c:436
#4  0xb6c1ffc0 in g_type_create_instance () from /usr/lib/arm-linux-gnueabihf/libgobject-2.0.so.0
#5  0xb6c0b7f2 in ?? () from /usr/lib/arm-linux-gnueabihf/libgobject-2.0.so.0
Backtrace stopped: previous frame identical to this frame (corrupt stack?)

when I add this line to the ini file (at the top): 
test = 1

I get a : 
Unable to write value: 12.000000 dB
Sequential loading of profile aborted.

?

pcercuei:
This should fix the issue: https://github.com/analogdevicesinc/linux/commit/9999669c5e49ee20a3a9ac8c7afc0cdb3bee1945

rgetz:
Does that mean I just need to update the kernel? (adi_update_boot)?

rgetz:
New kernel seems to work -- but it would be nice (but not necessary) to add the workaround in the osc app as well. (until the new kernel is uploaded on the web site).

Closing for now.

PR #581, opened 2026-09-18
Fix open CodeQL code scanning alerts

mhennerich:
## PR Description

Fixes all 10 currently open CodeQL code scanning alerts. One commit per area:

| Alert(s) | Rule | Commit |
|---|---|---|
| #4, #5, #6 | `cpp/command-line-injection` | **plugins: fmcomms2/5: Run fru-dump without a shell** |
| #7, #10 | `cpp/toctou-race-condition`, `cpp/wrong-type-format-argument` | **osc: Fix TOCTOU race and format argument in error message** |
| #8, #9 | `cpp/wrong-type-format-argument` | **fru: Fix format arguments in BOARD INFO error message** |
| #3 | `cpp/no-space-for-terminator` | **plugins: adrv9002: Allocate space for the string terminator** |
| #1, #2 | `actions/missing-workflow-permissions` | **ci: Restrict GITHUB_TOKEN permissions in workflows** |

Notes for review:

- **fru-dump**: the DCXO/XO "save to EEPROM" handlers built a command line with `sprintf()` and ran it through `popen()`. They now call a new `fru_dump_write_tuning()` helper in `eeprom.c`, which runs `fru-dump` via `posix_spawnp()` with the EEPROM path as a plain argument, so no shell is involved. `posix_spawnp()` rather than `fork()`/`exec()` so that no code runs in the child of this multi-threaded process.
  The realistic risk of the original code was low (the path comes from `find /sys -name eeprom`), but avoiding the shell removes the problem class entirely.
  **Behaviour change:** the old code closed the pipe without reading it, so `fru-dump` could be killed by SIGPIPE when printing and a successful write could be reported as "Error running fru-dump". Success is now decided purely by the exit status.
- **adrv9002**: this one was a real one-byte heap overflow (`malloc(strlen(s))` followed by copying the string including its terminator).
- **osc.c format argument (#10)**: most likely a false positive (the analysis doesn't see the libiio prototypes and assumes `int`); storing the name in a `const char *` first makes the type explicit.
- **Workflows**: neither workflow uses the token for anything beyond checkout, so `contents: read` is sufficient. CRLF line endings in `buildmingw.yml` are preserved.

Testing:

- All sources compile without warnings with `-Werror` against libiio v1.
- `fru_dump_write_tuning()` was compared against the old `popen()` code in a small harness (argument passing, non-zero exit, tool missing, path containing shell metacharacters) and run against the real `fru-dump` with invalid input. An actual EEPROM write on FMCOMMS hardware was **not** tested.
- osc was started against an ADALM-PLUTO (AD936X plugins load and work). The changed fmcomms EEPROM handlers are not reachable on that hardware.

The alerts should close once CodeQL rescans `main` after the merge.

## PR Type

## PR Checklist
- [x] I have commented new code, particulary complex or unclear areas 


dNechita (approved):
Looks ok to me.

